PR #22, opened 2025-10-19
feat: Initial Docling API and Client for health and source conversion

ThomasVitale:
This comes from the [Arconia Docling](https://docs.arconia.io/arconia/latest/integrations/docling/) library. It includes support for two endpoints:

* `/health`
* `/v1/convert/source`

The domain objects are modelled as Java records and Jackson-annotated, especially to configure the naming (though, Jackson is not required at runtime). There is a `DoclingApi` interface which would be the one implemented in Quarkus or Spring Boot. In the client package, there is a default implementation (`DoclingClient`) based on JDK HttpClient and Jackson.

I see this as a starting point, so to have something working to begin with. I'm open to evolve this in different ways moving forward.

* The API is represented by an interface and a collection of Java records and enums. Jackson annotations are used, but implementations are not required to use Jackson.
* The Client is a default implementation of the API. It uses the JDK HttpClient and Jackson. It also works as a reference implementation. Libraries and frameworks can implement the DoclingApi interface to provide their own implementation.
* Unit tests and integration tests are included for the most common paths.

### DIFF
--- a/api/build.gradle.kts
+++ b/api/build.gradle.kts
@@ -6,4 +6,5 @@ dependencies {
   implementation(platform(libs.jackson.bom))
 
   implementation("com.fasterxml.jackson.core:jackson-annotations")
+  testImplementation(project(":docling-testcontainers"))
 }
--- a/client/build.gradle.kts
+++ b/client/build.gradle.kts
@@ -9,5 +9,6 @@ dependencies {
   api(project(":docling-api"))
   api("tools.jackson.core:jackson-databind")
 
-  testImplementation("org.testcontainers:junit-jupiter")
+  testImplementation("org.testcontainers:testcontainers-junit-jupiter")
+  testImplementation(project(":docling-testcontainers"))
 }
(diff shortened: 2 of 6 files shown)


api/build.gradle.kts
```diff
@@ -2,3 +2,8 @@ plugins {
   id("docling-java-shared")
 }
 
+dependencies {
+  implementation(platform(libs.jackson.bom))
```
edeandrea, line 6: Shouldnt it be an API level dependency?

api/build.gradle.kts
```diff
@@ -2,3 +2,8 @@ plugins {
...
+
+  implementation("com.fasterxml.jackson.core:jackson-annotations")
```
edeandrea, line 8: Shouldn't it be an API level dependency?

edeandrea: Ahhh i see your note above about wanting to keep jackson out of the public api.

lordofthejars:
Sure, it is valid and probably this is what we need to do.moreover ifwe want to let implementators change it, then we are good to go.

edeandrea:
> @edeandrea thanks for the review! I like the idea of merging this and then look into the design we want to adopt moving forward.
> 
> I would suggest creating two separate issues where to discuss these subjects: one for the API and one for the Client. What do you think? I have some concerns about the interface and SPI approach for the domain APIs, but maybe it's better if I share them on a dedicated issue rather than here?

Done!

PR #27, opened 2025-10-23
feat: initial api & client implementation

### LINKED ISSUE #23: API Design
This is a parent issue for discussing the API design. We can create subtasks as needed.

_Originally posted by @ThomasVitale in https://github.com/docling-project/docling-java/issues/22#issuecomment-3427413526_

### LINKED ISSUE #24: Client design
This is a parent issue for discussing the client design. We can create subtasks as needed.

_Originally posted by @ThomasVitale in https://github.com/docling-project/docling-java/issues/22#issuecomment-3427413526_

edeandrea:
An idea for api design. This still allows using Java records to implement the model objects, but also makes these objects backwards-compatible by introducing interfaces & builders.

**NOTE** I've only applied the pattern to the `DocumentResponse` so that we can decide if this is a good approach or not. If so, we can apply this pattern to the others.

FIxes #23
Fixes #24

### DIFF
--- a/docs/build.gradle.kts
+++ b/docs/build.gradle.kts
@@ -22,7 +22,7 @@ mkdocs {
       "project-version" to "${project.version}",
       "project-groupId" to "${project.group}",
       "project-artifactId" to "${rootProject.name}",
-      "api-artifactId" to project(":docling-serve-api").name,
+      "serve-api-artifactId" to project(":docling-serve-api").name,
       "serve-client-artifactId" to project(":docling-serve-client").name,
       "testing-artifactId" to project(":docling-testing").name,
       "testcontainers-artifactId" to project(":docling-testcontainers").name


edeandrea:
@ThomasVitale @lordofthejars take a look and let me know what you think of this approach. I didn't get into ServiceLoaders or anything like that. I think that overcomplicates things.

client/src/main/java/ai/docling/client/DoclingClient.java
```diff
@@ -14,30 +14,35 @@
 import ai.docling.api.convert.response.ConvertDocumentResponse;
 import ai.docling.api.health.HealthCheckResponse;
 
+import static ai.docling.api.util.ValidationUtils.*;
```
lordofthejars: Do we want to use the star import? I am not saying it isbad I am just wondering

edeandrea: Usually for `*` static imports my general rule of thumb (for myself at least) is that if its an internal class local to the codebase, then its ok, but if its pulling in 3rd party class/library then I tend to not import static `*`. Since `ValidationUtils` is part of the docling-java codebase (& It'll be used a lot) I tend to `import static *`.

I generally don't do any regular `import *`, only some `import static *`.

But I don't have a severely strong opinion on things. If @ThomasVitale @lordofthejars prefer never using `import static *` then thats fine too.

lordofthejars: It is good in my opinion, I was just raising for just in case IDE collapse the imports and you didn't notice

ThomasVitale: I generally avoid star imports, but I guess there's some rule in the `.editorconfig` file forcing them because I noticed that in my previous PR I got some star imports automatically applied, removing my explicit ones.

I'm fine either way, as long as we're consistent and we can automate the enforcement of whatever rule we decide to be compliant with.

edeandrea: I'm fine to remove `*` imports. I think the `.editorconfig` file was just one that I copied from somewhere else, so we can certainly do what we'd like.

I usually like seeing static imports _before_ other imports too, so maybe we need to make a few adjustments.

edeandrea:
@ThomasVitale @lordofthejars I've found a way to support both Jackson 2 & 3 in the API & client

See https://github.com/docling-project/docling-java/pull/27/commits/8f88ebee8db236b90036390dc7622c61a1f2400a and https://github.com/docling-project/docling-java/pull/27/commits/1b3c2d93e0629e98901e1d194fc080d7131ed3b8

Since they are annotations, I can have both Jackson 2 & 3 at the API level. Then at the client level I can support both 2 & 3 through a Factory interface, which can detect which client to give based on whats on the classpath.

api/src/main/java/ai/docling/api/convert/response/DocumentResponse.java
```diff
@@ -2,33 +2,271 @@
...
 
-@JsonInclude(JsonInclude.Include.NON_NULL)
-public record DocumentResponse(
+import tools.jackson.databind.annotation.JsonDeserialize;
+import tools.jackson.databind.annotation.JsonPOJOBuilder;
 
-    @JsonProperty("doctags_content") @Nullable String doctagsContent,
+@JsonInclude(JsonInclude.Include.NON_ABSENT)
+@JsonDeserialize(builder = DocumentResponse.Builder.class)
+@com.fasterxml.jackson.databind.annotation.JsonDeserialize(builder = DocumentResponse.Builder.class)
```
pjfanning: this line needs to match line 12 above - and it doesn't - probably best not to use FQCN here

edeandrea: I'm not sure what you mean? Since the class names (`JsonDeserialize`) are the same, wouldn't you need FQCN to differentiate/have both on the same class?

pjfanning: Ok. I get what you trying to now. I didn't spot that you were adding both Jackson 2 and 3 annotations to the same method.
I suspect that this will be a problem as it might stop you from making the 2 jackson jars optional. This might require that you have both at runtime.

In your post on the Jackson project, I thought you were talking about having Jackson 2 and 3 code in your lib but I assumed that there would be separate classes for the Jackson 2 cases and for the Jackson 3 cases.

In the end, you can force your users to have both versions of Jackson on the classpath but users in constrained envs like Android use cases might not like being forced to have extra jars on the classpath.

It might be that Java is less strict at runtime about having classes present when they relate to annotations (but I would suspect that this is not the case).

edeandrea: Actually when it comes to annotations, if the class isn't on the classpath [Java will simply ignore it](https://www.baeldung.com/classnotfoundexception-missing-annotation). Thats the magic that makes this work :)

lordofthejars:
I am not a big fan of Lombok, but given the nature of the project I'd say taht everything that can help us on less maintainance  is well welcomed.

lordofthejars:
I'd say go ahead, let's keep things simple for now, moreover, I thikn we should monitor the number of downloads and if we see there is a massive adoption, then, let's figure out to make things more beauty

ThomasVitale:
I'm not a fan of Lombok either, but I understand the context and see the value in this case :)

edeandrea:
@ThomasVitale are you good with this approach and everything I've done here?

.github/workflows/build.yml
```diff
@@ -28,10 +28,10 @@ jobs:
           - 21
           - 25
         module:
-          - api
-          - client
-          - testing
-          - testcontainers
+          - docling-api
```
ThomasVitale: Since the idea is to possibly have a `docling-client` module in the future using Docling directly (no Docling Serve like in `docling-serve-client`), I wonder if we should similarly have a `docling-serve-api` module (for all the code currently in docling-api) and a `docling-api` module that would only include the Docling core APIs (e.g. `Document`). In alternative, single module, but two sub packages (`core`, `serve`)?

edeandrea: Sorry I've totally changed this comment after thinking about it a bit.

That might not be a bad idea. We can just rename docling-api to docling-serve-api for now. We can build a placeholder for docling-api (or introduce it later on).

edeandrea: @ThomasVitale i've gone ahead and restructured this. I created a new top-level folder `docling-serve` and moved both `docling-serve-api` (formerly `docling-api`) and `docling-serve-client` into this folder.

I also changed the package structure for `docling-serve-api` (formerly `docling-api`) to `ai.docling.api.serve` (formerly `ai.docling.api`).

See https://github.com/docling-project/docling-java/pull/27/commits/dfa67558366a70bfb7e7256fb53ab65bbdd41fda

edeandrea:
@ThomasVitale are you ok if I go ahead and merge this? Since it involves directory/name changes I'd like to get this in as I'm working on other things (version tester, release automation, etc) that will be impacted by this.

We can always improve it as we move along.

ThomasVitale:
Yes, let's merge this!

ThomasVitale (approved):

PR #92, opened 2025-11-23
chore(deps): Update Gradle Wrapper 9.2.0 to 9.2.1

ThomasVitale:
- Update Gradle Wrapper to 9.2.1 and add checksum for validation
- Refine GitHub Workflows to leverage the Gradle caching from the Setup Gradle action instead of Setup Java
- Optimize Git checkout operations in GitHub Workflows so that only the current branch is downloaded.

From the Setup Gradle actions [docs](https://github.com/gradle/actions/blob/main/docs/setup-gradle.md):

> More sophisticated and more efficient caching of Gradle User Home between invocations, compared to setup-java and most custom configurations using actions/cache. 

> Detailed reporting of cache usage and cache configuration options allow you to [optimize the use of the GitHub actions cache](https://github.com/docling-project/docling-java/pull/92#optimizing-cache-effectiveness).


ThomasVitale:
The Setup Gradle action is actually already performing the Gradle Wrapper validation (see [docs](https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#gradle-wrapper-validation)), so we could remove the dedicated workflow. For now I kept it.

> By default, this action will perform the same wrapper validation as is performed by the dedicated [wrapper-validation action](https://github.com/gradle/actions/blob/main/docs/wrapper-validation.md). This means that invalid wrapper jars will be automatically detected when using setup-gradle.

gradle.properties
```diff
@@ -1 +1,3 @@
+org.gradle.jvmargs=-Xmx8g
```
ThomasVitale: This is useful if running, for example, OpenRewrite recipes which need lots of resources to parse the project.

edeandrea: Should this setting be applied globally for every user who wants to run the project though?

ThomasVitale: Hmm good point! It could have downsides on more constrained environments. Let me remove this.

ThomasVitale: Removed

ThomasVitale:
I removed the "Gradle Wrapper Validation" workflow as a mandatory check for PRs in the GitHub settings, since the workflow has been removed now.

Issue #98, opened 2025-11-24
Add DoclingDocument core API


ThomasVitale:
To enable downstream use cases for RAG and other GenAI applications, the DoclingDocument API is an essential part.

Python type: https://github.com/docling-project/docling-core/blob/main/docling_core/types/doc/document.py

PR #99, opened 2025-11-24
feat: Add DoclingDocument API in new docling-core module

ThomasVitale:
Fixes gh-98

---

I tried to map the structure of the Docling project, which uses a dedicated Docling Core package for data types models: https://github.com/docling-project/docling-core. So I introduced a new `docling-core` module, including some basic documentation. Since we are using the `ai.docling.api.serve` and `ai.docling.client.serve` packages for the other two modules, I went with `ai.docling.api.core` for consistency. But I'm open to other solutions (e.g. `ai.docling.core`, `ai.docling.serve.api` and `ai.docling.serve.client`).

I have added an explicit `DoclingDocument` class following the same approach used for the Serve APIs: Lombok + Jackson 2 and 3 compatibility. This is probably gonna be one of the hardest classes to maintain manually, but until we come up with a reliable automated process, I guess we'll have to. I initially thought about making it a bit more generic (e.g. using simple Strings rather than enums), but that would reduce its usefulness by a lot. In order to make it useful for downstream scenarios (e.g. integrations with LangChain4j or Spring AI for RAG and agent workflows), we need to have a full type-safe representation.

Finally, I considered how to switch from `Map` to `DoclingDocument` in a backward compatible way in the context of `DocumentResponse`, but in the end I thought it wasn't worth the effort considering we're at very beginning and using the returned Map is pointless without a proper data type. So, I switched `DocumentResponse.getJsonContent()` to return a `DoclingDocument` instead of `Map<String, Object>`.

Thoughts? @edeandrea @lordofthejars

### DIFF
--- a/docling-testing/docling-version-tests/build.gradle.kts
+++ b/docling-testing/docling-version-tests/build.gradle.kts
@@ -15,7 +15,6 @@ dependencies {
   implementation(libs.semver4j)
   implementation(libs.assertj.core)
   implementation(libs.quarkus.github.api)
-  implementation(project(":docling-core"))
   implementation(project(":docling-serve-client"))
   implementation(project(":docling-testcontainers"))
 
--- a/docs/src/doc/docs/index.md
+++ b/docs/src/doc/docs/index.md
@@ -22,7 +22,7 @@ Docling Java is the official Java client and tooling for [Docling](https://docli
 
 This repository provides a set of artifacts you can mix and match depending on your needs:
 
-- [`docling-core`](core.md): Java API for working with the data types used by Docling for document representation.
+- [`docling-core`](core.md): Java API for working with the data types used by Docling for document representation (see [Docling Core](https://github.com/docling-project/docling-core)).
 - [`docling-serve-api`](docling-serve/serve-api.md): Java API for interacting with a [Docling Serve](https://github.com/docling-project/docling-serve) backend (framework‑agnostic).
 - [`docling-serve-client`](docling-serve/serve-client.md): Reference HTTP client built with Java `HttpClient` and [Jackson](https://github.com/FasterXML/jackson) for connecting to a [Docling Serve](https://github.com/docling-project/docling-serve) endpoint.
 - [`docling-testing`](testing.md): Utilities for testing Docling integrations in your codebase.


docling-testing/docling-version-tests/build.gradle.kts
```diff
@@ -15,6 +15,7 @@ dependencies {
   implementation(libs.semver4j)
   implementation(libs.assertj.core)
   implementation(libs.quarkus.github.api)
+  implementation(project(":docling-core"))
```
edeandrea: Is this really needed? Wouldn't the dependency on the `:docling-serve-client` bring it in transitively?

docs/src/doc/docs/index.md
```diff
@@ -22,7 +22,8 @@ Docling Java is the official Java client and tooling for [Docling](https://docli
 
 This repository provides a set of artifacts you can mix and match depending on your needs:
 
-- [`docling-serve-api`](docling-serve/serve-api.md): Core Java API for interacting with a [Docling Serve](https://github.com/docling-project/docling-serve) backend (framework‑agnostic).
+- [`docling-core`](core.md): Java API for working with the data types used by Docling for document representation.
```
edeandrea: Should we link to the docling document repo directly here too?

edeandrea:
Thanks @ThomasVitale!

> Finally, I considered how to switch from Map to DoclingDocument in a backward compatible way in the context of DocumentResponse, but in the end I thought it wasn't worth the effort considering we're at very beginning and using the returned Map is pointless without a proper data type. So, I switched DocumentResponse.getJsonContent() to return a DoclingDocument instead of Map<String, Object>.

I agree. I think we can still make breaking changes for a bit.

I'm also going to be OOO the rest of this week, so don't let me be a bottleneck if you are planning to work on this. I trust your judgement - feel free to merge things.

docling-testing/docling-version-tests/build.gradle.kts
ThomasVitale: Thanks for mentioning this. It's not actually needed. Leftover from my initial test where the client module was not brining in the dependency transitively. I removed it now.

docs/src/doc/docs/index.md
ThomasVitale: Yes, good idea. I added a link.

Issue #105, opened 2025-11-25
Package renaming


edeandrea:
### Discussed in https://github.com/docling-project/docling-java/discussions/103

<div type='discussions-op-text'>

<sup>Originally posted by **edeandrea** November 25, 2025</sup>
I was thinking about the package naming a bit (for docling-serve).

Since we now have a core module, should we rename `ai.docling.api.serve` to just `ai.docling.serve` so that the hierarchy is consistent with the other modules (`ai.docling.core`/`ai.docling.testcontainers`)?</div>

PR #107, opened 2025-11-26
refactor: Rename packages for Docling Serve modules

ThomasVitale:
* The classes in the `docling-serve-api` module have been moved from the `ai.docling.api.serve` package to the `ai.docling.serve.api` package.
* The classes in the `docling-serve-client` module have been moved from the `ai.docling.client.serve` package to the `ai.docling.serve.client` package.
* A new 'What's New' page has been added to include highlights from each release.
* A new 'Upgrade' page has been added to include information on how to update automatically with OpenRewrite. For now, I placed the recipe in the Migrations project I have already set up for recipes, but int he future we might consider setting something up under the Docling organization. https://github.com/arconia-io/arconia-migrations/commit/25ea03b1e2ea04584021676b7a3b8465d04ee487

Fixes gh-105


docs/src/doc/docs/upgrading.md
edeandrea, line 1: Does this really belong here or did this get copied over by mistake from Arconia? Seems weird to redirect the user to Arconia to upgrade docling-java?

ThomasVitale: I started there temporarily to validate the solution since I had an OpenRewrite module already and could publish the Docling Recipe, but my plan was to set up a "rewrite-docling" project. Too ambitious though in the context of one PR since I didn't have time to complete it yet :) I removed this page and will create a separate issue, so we don't delay the other changes.

edeandrea: You answered my next question :) Its a great idea. Maybe we call the project `docling-update`, although I'm not married to the name.

PR #204, opened 2025-12-12
feat: Add Async support

### LINKED ISSUE #127: Add async support
Add async support for document conversion

oscerd:
This should fix #127


edeandrea:
I'll take a look on monday (I'm away from the computer today and this weekend)

oscerd:
No worries. Enjoy the weekend!

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/DoclingServeClient.java
```diff
@@ -329,6 +420,8 @@ public abstract static class DoclingServeClientBuilder<C extends DoclingServeCli
     private boolean logRequests = false;
     private boolean logResponses = false;
     private boolean prettyPrintJson = false;
+    private long asyncPollIntervalMs = 2000; // Default: poll every 2 seconds
```
ThomasVitale: For these two fields, you could consider using a `Duration` instead of a `long` to increase the readability, simplify the validation, and also align with other timeout-related values in the Serve API for which we used a `Duration`. What do you think?

oscerd: You're right. I'll double check.

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeApi.java
```diff
@@ -1,10 +1,58 @@
...
+import ai.docling.serve.api.async.response.TaskResponse;
+import ai.docling.serve.api.convert.request.ConvertDocumentRequest;
+import ai.docling.serve.api.convert.response.ConvertDocumentResponse;
+
 /**
  * Docling Serve API interface.
  */
 public interface DoclingServeApi
     extends DoclingServeHealthApi, DoclingServeConvertApi, DoclingServeChunkApi, DoclingServeClearApi, DoclingServeTaskApi {
+  /**
```
ThomasVitale: After the recent refactoring that @edeandrea did, should these new APIs be added to `DoclingServeConvertApi` rather than here directly?

ThomasVitale: Or do we want a `DoclingServeAsyncConvertApi`?

ThomasVitale: In any case, it would be great if you could mention these new operations in the _What's new_ page of the documentation, so that they're highlighted when we build the next release: https://github.com/docling-project/docling-java/blob/main/docs/src/doc/docs/whats-new.md

edeandrea: > After the recent refactoring that @edeandrea did, should these new APIs be added to `DoclingServeConvertApi` rather than here directly?

If you look at the Docling Serve documentation they break it down into the health, convert, chunk, clear, and task apis. I think we should keep the operations consistent to where docling serve keeps them. They mix the blocking & async operations within those.

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/convert/response/ConversionStatus.java
edeandrea, line 1: These response objects probably belong in the convert/response package along with the rest of the convert response objects?

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeApi.java
```diff
@@ -1,10 +1,58 @@
...
+   */
+  TaskResponse convertSourceAsync(ConvertDocumentRequest request);
+
+  /**
+   * Checks the status of an asynchronous conversion task.
+   *
+   * @param taskId the task ID returned from {@link #convertSourceAsync(ConvertDocumentRequest)}
+   * @return a {@link ConversionStatus} object with the current status of the task
+   */
+  ConversionStatus checkConversionStatus(String taskId);
```
edeandrea: I was thinking that these async apis would return `CompletableFuture`? That way the implementation would take care of all the polling necessary?

And also clients could then chain with things like `.andThen()`/`.andThenAsync()`?

That way also downstream frameworks like Quarkus & Spring boot could re-implement with their own async apis (Mutiny/Webflux)?

We also need to be aware of thread management. The polling mechanisms would have to happen on another thread with the ability for customization.

@ThomasVitale thoughts?

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeApi.java
oscerd: I'll move the Convert operation into the DoclingServeConvertApi.

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/convert/response/ConversionStatus.java
oscerd: +1

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeApi.java
oscerd: In Camel the CompletableFuture was used only in the polling operation, so I moved that logic as it was, but I'll rework this.

edeandrea: We can let @ThomasVitale weigh in with his thoughts about `CompletableFuture` if you'd like. It was just my thought. 

Either way though the polling for completion has to happen on a different thread so that the thread that launches the request does not block.

oscerd: I'm trying to move the entire implementation to CompletableFuture.

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/convert/response/ConversionStatus.java
oscerd: Moved.

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/DoclingServeClient.java
oscerd: I changed the implementation for CompletableFuture, but I need to get back to duration.

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeApi.java
ThomasVitale: I agree on the usage of `CompletableFuture` as it would lay a solid foundation for frameworks/apps to build on top of it and configure thread management and integrations with async or reactive frameworks.

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/operations/ConvertOperations.java
```diff
@@ -1,29 +1,111 @@
...
+  @Override
+  public CompletableFuture<ConvertDocumentResponse> convertSourceAsyncAndWait(ConvertDocumentRequest request, Executor executor) {
+    return this.asyncOperations.convertSourceAsyncAndWait(request, executor);
+  }
+
+  /**
+   * Interface for async conversion operations. This allows the main client to provide
+   * the async implementation while ConvertOperations handles the delegation.
+   */
+  public interface AsyncConvertOperations {
```
oscerd: This is just a way of having an Async implementation with the ConvertOperation delegating. It's a way of doing this, but I'm open to other solution

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeConvertApi.java
```diff
@@ -9,11 +14,91 @@
...
+  CompletableFuture<TaskResponse> convertSourceAsync(ConvertDocumentRequest request);
+
+  /**
+   * Checks the status of an asynchronous conversion task.
+   *
+   * @param taskId the task ID returned from {@link #convertSourceAsync(ConvertDocumentRequest)}
+   * @return a {@link CompletableFuture} that completes with a {@link ConversionStatus} object
+   *         containing the current status of the task
+   */
+  CompletableFuture<ConversionStatus> checkConversionStatus(String taskId);
```
edeandrea: I don't think we need this? The task api is already implemented in `ai.docling.serve.api.DoclingServeTaskApi`

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeConvertApi.java
```diff
@@ -9,11 +14,91 @@
...
+   *
+   * <p>This method starts a background conversion task and immediately returns a
+   * {@link CompletableFuture} that completes with a task ID that can be used to poll
+   * the status and retrieve results when the conversion completes.
+   *
+   * @param request the {@link ConvertDocumentRequest} containing the source(s) and conversion options.
+   * @return a {@link CompletableFuture} that completes with a {@link TaskResponse} containing
+   *         the task ID for tracking the async operation.
+   */
+  CompletableFuture<TaskResponse> convertSourceAsync(ConvertDocumentRequest request);
```
edeandrea: Shouldn't this return `CompletableFuture<ConvertDocumentResponse>`? I would think its implementation would handle all the necessary polling (via the task API) and then return the completed future once it is complete?

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeConvertApi.java
```diff
@@ -9,11 +14,91 @@
...
+  CompletableFuture<ConversionStatus> checkConversionStatus(String taskId);
+
+  /**
+   * Retrieves the result of a completed asynchronous conversion task.
+   *
+   * @param taskId the task ID returned from {@link #convertSourceAsync(ConvertDocumentRequest)}
+   * @return a {@link CompletableFuture} that completes with a {@link ConvertDocumentResponse}
+   *         containing the conversion result
+   */
+  CompletableFuture<ConvertDocumentResponse> getConversionResult(String taskId);
```
edeandrea: I don't think this is needed?

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeConvertApi.java
```diff
@@ -9,11 +14,91 @@
...
+   *     .thenAccept(markdown -> System.out.println(markdown))
+   *     .exceptionally(ex -> { ex.printStackTrace(); return null; });
+   * }</pre>
+   *
+   * @param request the {@link ConvertDocumentRequest} containing the source(s) and conversion options.
+   * @return a {@link CompletableFuture} that completes with the {@link ConvertDocumentResponse}
+   *         when the conversion is finished, or completes exceptionally if the conversion fails
+   *         or times out.
+   */
+  CompletableFuture<ConvertDocumentResponse> convertSourceAsyncAndWait(ConvertDocumentRequest request);
```
edeandrea: I don't think we need an `asyncAndWait`? If users want that they could just do `CompletableFuture.supplyAsync(() -> convert(request))`?

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeConvertApi.java
```diff
@@ -9,11 +14,91 @@
...
+   *     .join();
+   * }</pre>
+   *
+   * @param request the {@link ConvertDocumentRequest} containing the source(s) and conversion options.
+   * @param executor the {@link Executor} to use for polling operations.
+   * @return a {@link CompletableFuture} that completes with the {@link ConvertDocumentResponse}
+   *         when the conversion is finished, or completes exceptionally if the conversion fails
+   *         or times out.
+   */
+  CompletableFuture<ConvertDocumentResponse> convertSourceAsyncAndWait(ConvertDocumentRequest request, Executor executor);
```
edeandrea: I don't think we need an `asyncAndWait`? If users want that they could just do `CompletableFuture.supplyAsync(() -> convert(request), Executor)`?

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/operations/ConvertOperations.java
edeandrea: I don't think you need this? The Task api is already implemented. You could just add `TaskOperations` to the constructor of `ConvertOperations`.

See my other comments. I think you only need 1 additional (public) method in the `ConvertOperations` class

```java
public CompletableFuture<ConvertDocumentResponse> convertSourceAsync(ConvertDocumentRequest request)
````

Then in the implementation you could use the task api to poll, and then complete the future once the task has completed.

**Comment by oscerd** [comment]
It should be better now. I spent an hour or so looking at the API a bit deeper.

**Review by edeandrea — DISMISSED** [review]
Looks good! I hope you don't mind but I made a few tweaks. `CompletableFuture` has built-in polling capabilities, so you don't need to create your own polling executor. (see https://github.com/docling-project/docling-java/pull/204/commits/f5033ac92dcdc7c0c9d65b2337b28d3d3aa96805)

I also added the polling properties to the `DoclingServeApi` builder so it can be configured by any client implementing the api.

@oscerd @ThomasVitale Have a look and let me know what you think.

File: docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeConvertApi.java [hunk 10]
```diff
@@ -9,11 +11,35 @@
...
    * @return a {@link ConvertDocumentResponse} containing the processed document data, processing details, and any errors.
    */
   ConvertDocumentResponse convertSource(ConvertDocumentRequest request);
+
+  /**
+   * Initiates an asynchronous conversion of the provided document source(s) and returns a
+   * {@link CompletableFuture} that completes when the conversion is done.
+   *
+   * <p>This method starts the conversion, polls the status in the background, and completes
+   * the future with the result when the conversion finishes. The polling happens on a separate
```
ThomasVitale: Perhaps we should remove this final sentence as it's up to the framework/client to configure how the polling happens (platform thread, virtual thread, some other mechanism provided by Mutiny/Reactor). Thoughts?

docs/src/doc/docs/whats-new.md
```diff
@@ -10,6 +10,7 @@ Docling Java {{ gradle.project_version }} provides a number of new features, enh
 * Adding `pretty-print` configuration option to `DoclingServeClient` to enable pretty printing of JSON requests and responses.
 * Adding the ability to specify an api key for the Docling Serve Testcontainer.
 * Adding the ability to specify the api key for the Docling Serve requests.
+* Adding async conversion support via `convertSourceAsync(ConvertDocumentRequest)` which returns a `CompletableFuture<ConvertDocumentResponse>`. The method handles all polling internally and completes the future when the conversion is done. Configurable via `asyncPollInterval(Duration)` and `asyncTimeout(Duration)` builder methods.
```
ThomasVitale, line 13:
```suggestion
* Adding async conversion support in `DoclingServeApi` via `convertSourceAsync(ConvertDocumentRequest)` which returns a `CompletableFuture<ConvertDocumentResponse>`. The method handles all polling internally and completes the future when the conversion is done. Configurable via `asyncPollInterval(Duration)` and `asyncTimeout(Duration)` builder methods.
```

oscerd: +1

oscerd: +1

ThomasVitale:
Looks good to me! Just left a couple of comments about the docs. I like the clean API returning `CompletableFuture` and allowing clients/frameworks to adapt it to different asynchronous setups.

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeConvertApi.java
edeandrea: Since this is the API then yes we should probably leave out anything that describes the "how". Its up to the implementation of the api to decide that for themselves.

Now, if we want to add a comment to the `docling-serve-client` docs about how _it_ does it, then that would surely be good, because its specific to that implementation.

edeandrea (dismissed):

edeandrea:
@all-contributors add @oscerd as code docs

ThomasVitale (dismissed):

ThomasVitale:
Great, thank you both! This is a very useful addition to the project.

PR #229, opened 2025-12-16
feat: Add file-based operations for conversion and chunking

### LINKED ISSUE #227: Implement process file operations
Implement the following APIs:

- Process File
- Process File async
- Chunk Files With Hybridchunker
- Chunk Files With Hybridchunker async
- Chunk Files With Hierarchicalchunker
- Chunk Files With Hierarchicalchunker async

edeandrea:
- Introduced support for file-based conversion and chunking workflows.
- Added synchronous and asynchronous methods for handling file paths in API and client layers.
- Enhanced error handling and validation for file-based operations.
- Updated tests to cover new file-based use cases, including edge cases and null checks.

Fixes #227


edeandrea:
So after some examination, really the only difference between "process/chunk file" and "process/chunk url" is that the process/chunk file uses `multipart/form-data`.

We can do the same thing that `/v1/convert/file` does using `/v1/convert/source` by sending the base64 encoded files in the request body.

So all I really did here was add some convenience methods to `DoclingServeConvertApi` and `DoclingServeChunkApi` to make constructing the request objects & reading the files easier.

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeConvertApi.java
```diff
@@ -35,9 +66,51 @@ public interface DoclingServeConvertApi {
    * }</pre>
    *
    * @param request the {@link ConvertDocumentRequest} containing the source(s) and conversion options.
-   * @return a {@link CompletableFuture} that completes with the {@link ConvertDocumentResponse}
+   * @return a {@link CompletionStage} that completes with the {@link ConvertDocumentResponse}
    *         when the conversion is finished, or completes exceptionally if the conversion fails
    *         or times out.
    */
-  CompletableFuture<ConvertDocumentResponse> convertSourceAsync(ConvertDocumentRequest request);
+  CompletionStage<ConvertDocumentResponse> convertSourceAsync(ConvertDocumentRequest request);
```
ThomasVitale, line 73: Good idea using `CompletionStage` over `CompletableFuture`! It will make it more flexible to adapt to reactive libraries.

edeandrea: Yeah I saw that when I was in there doing this PR and thought making it more generic was better.

PR #277, opened 2026-01-14
fix: string.isBlank() is equivalent to string.trim().isBlank()

oscerd:
isBlank() already handles whitespace-only strings directly without creating an intermediate object (starting from Java 11+). The trim() call was redundant - it would only matter if we cared about the trimmed result, but we're just checking blankness.


oscerd:
This was really a stupid one, but it makes sense, we won't allocate a  new trimmed String object

PR #317, opened 2026-02-03
feat: Support configuring timeouts in DoclingServeApi.Builder

ThomasVitale:
Add support for connect and read timeout in DoclingServeApi.Builder and implement the new fields in the reference client implementation.

### DIFF
--- a/docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/DoclingServeClient.java
+++ b/docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/DoclingServeClient.java
@@ -43,6 +43,7 @@ import ai.docling.serve.api.task.request.TaskResultRequest;
 import ai.docling.serve.api.task.request.TaskStatusPollRequest;
 import ai.docling.serve.api.task.response.TaskStatusPollResponse;
 import ai.docling.serve.api.util.Utils;
+import ai.docling.serve.api.util.ValidationUtils;
 import ai.docling.serve.api.validation.ValidationError;
 import ai.docling.serve.api.validation.ValidationException;
 import ai.docling.serve.client.operations.ChunkOperations;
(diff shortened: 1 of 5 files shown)


docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/DoclingServeClient.java
```diff
@@ -467,6 +479,20 @@ public B prettyPrint(boolean prettyPrint) {
       return (B) this;
     }
 
+    @Override
+    public B connectTimeout(Duration connectTimeout) {
+      ValidationUtils.ensurePositiveDuration(connectTimeout, "connectTimeout");
```
edeandrea: Would it be better to do the validation in the constructor that the build method calls rather than here in the builder?

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/DoclingServeClient.java
```diff
@@ -467,6 +479,20 @@ public B prettyPrint(boolean prettyPrint) {
...
+      this.connectTimeout = connectTimeout;
+      return (B) this;
+    }
+
+    @Override
+    public B readTimeout(Duration readTimeout) {
+      ValidationUtils.ensurePositiveDuration(readTimeout, "readTimeout");
```
edeandrea: Would it be better to do the validation in the constructor that the build method calls rather than here in the builder?

ThomasVitale: Good idea, I'll move the validation

ThomasVitale: Done

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/DoclingServeClient.java
ThomasVitale: Done

PR #329, opened 2026-02-06
feat(client): implement docling-serve-grpc service

### LINKED ISSUE #325: Integrate a streaming gRPC client
This originated by an idea from @krickert in https://github.com/quarkiverse/quarkus-docling/issues/95

@ThomasVitale what do you think? Basically create a protobuf schema around what we have and expose a gRPC interface that talks to the docling-serve api?

If we do it I think we'd introduce a new `docling-serve-api-grpc` module or something like that.

krickert:
This pull request introduces a professional gRPC client and service bridge for Docling Serve, designed for enterprise-grade pipeline integrations and long-term archival storage.

### Key Enhancements

- **Professional gRPC API Design:** Implements the `ai.docling.serve.v1` contract with strict versioning throughout both Protobuf and Java packages.
- **Unique Request/Response Wrappers:** Adheres to Google API Design and Buf linting standards by using dedicated wrapper messages for every RPC, ensuring future-proof method independence.
- **Robust Null-Safety:** Implements exhaustive defensive mapping for the entire Docling document schema, guarding all collections, boxed primitives, and enums.
- **Server-Side Polling (Watch Pattern):** Introduces "Watch" RPCs that manage asynchronous task monitoring on the server side, streaming status updates back to the client.
- **Comprehensive Verification:** Includes unit, bidirectional mapping, and Testcontainers-based integration suites (all 78 tests passing).
- **JPMS Support:** Fully compatible with the Java Platform Module System.

Closes #325

### DIFF
--- a/docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeChunkApi.java
+++ b/docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeChunkApi.java
@@ -9,7 +9,6 @@ import ai.docling.serve.api.chunk.request.ChunkType;
 import ai.docling.serve.api.chunk.request.HierarchicalChunkDocumentRequest;
 import ai.docling.serve.api.chunk.request.HybridChunkDocumentRequest;
 import ai.docling.serve.api.chunk.response.ChunkDocumentResponse;
-import ai.docling.serve.api.task.response.TaskStatusPollResponse;
 import ai.docling.serve.api.util.FileUtils;
 import ai.docling.serve.api.util.ValidationUtils;
 
@@ -27,17 +26,6 @@ public interface DoclingServeChunkApi {
    * @throws ai.docling.serve.api.validation.ValidationException If request validation fails for any reason.
    */
   ChunkDocumentResponse chunkSourceWithHierarchicalChunker(HierarchicalChunkDocumentRequest request);
-
-  /**
-   * Submits an asynchronous hierarchical chunking task and returns the initial task status
-   * immediately without waiting for completion. The returned {@link TaskStatusPollResponse}
-   * contains the task ID that can be used to track progress and retrieve results.
-   *
-   * @param request the request containing the document source(s) and options for hierarchical chunking
-   * @return a {@link TaskStatusPollResponse} containing the task ID and initial status
-   * @throws ai.docling.serve.api.validation.ValidationException If request validation fails for any reason.
-   */
-  TaskStatusPollResponse submitChunkHierarchicalSource(HierarchicalChunkDocumentRequest request);
 
   /**
    * Processes and chunks the specified files into smaller, structured pieces
@@ -97,18 +85,6 @@ public interface DoclingServeChunkApi {
    * @throws ai.docling.serve.api.validation.ValidationException If request validation fails for any reason.
    */
   ChunkDocumentResponse chunkSourceWithHybridChunker(HybridChunkDocumentRequest request);
-
-  /**
-   * Submits an asynchronous hybrid chunking task and returns the initial task status
-   * immediately without waiting for completion. The returned {@link TaskStatusPollResponse}
-   * contains the task ID that can be used to track progress and retrieve results.
-   *
-   * @param request the request containing the document source(s), options for conversion,
-   *                and hybrid chunking parameters
-   * @return a {@link TaskStatusPollResponse} containing the task ID and initial status
-   * @throws ai.docling.serve.api.validation.ValidationException If request validation fails for any reason.
-   */
-  TaskStatusPollResponse submitChunkHybridSource(HybridChunkDocumentRequest request);
 
   /**
    * Processes and chunks the specified files into smaller, structured pieces
--- a/docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeConvertApi.java
+++ b/docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeConvertApi.java
@@ -8,7 +8,6 @@ import org.jspecify.annotations.Nullable;
 
 import ai.docling.serve.api.convert.request.ConvertDocumentRequest;
 import ai.docling.serve.api.convert.response.ConvertDocumentResponse;
-import ai.docling.serve.api.task.response.TaskStatusPollResponse;
 import ai.docling.serve.api.util.FileUtils;
 import ai.docling.serve.api.util.ValidationUtils;
 
@@ -26,19 +25,6 @@ public interface DoclingServeConvertApi {
    * @throws ai.docling.serve.api.validation.ValidationException If request validation fails for any reason.
    */
   ConvertDocumentResponse convertSource(ConvertDocumentRequest request);
-
-  /**
-   * Submits an asynchronous conversion of the provided document source(s) and returns
-   * the initial task status immediately without waiting for completion. The returned
-   * {@link TaskStatusPollResponse} contains the task ID that can be used with
-   * {@link DoclingServeTaskApi#pollTaskStatus} and {@link DoclingServeTaskApi#convertTaskResult}
-   * to track progress and retrieve results.
-   *
-   * @param request the {@link ConvertDocumentRequest} containing the source(s) and conversion options.
-   * @return a {@link TaskStatusPollResponse} containing the task ID and initial status.
-   * @throws ai.docling.serve.api.validation.ValidationException If request validation fails for any reason.
-   */
-  TaskStatusPollResponse submitConvertSource(ConvertDocumentRequest request);
 
   /**
    * Converts the specified files into a processed document using default options.
--- a/docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/DoclingServeClient.java
+++ b/docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/DoclingServeClient.java
@@ -300,28 +300,13 @@ public abstract class DoclingServeClient extends HttpOperations implements Docli
   }
 
   @Override
-  public TaskStatusPollResponse submitConvertSource(ConvertDocumentRequest request) {
-    return this.convertOps.submitConvertSource(request);
-  }
-
-  @Override
   public ChunkDocumentResponse chunkSourceWithHierarchicalChunker(HierarchicalChunkDocumentRequest request) {
     return this.chunkOps.chunkSourceWithHierarchicalChunker(request);
   }
 
   @Override
-  public TaskStatusPollResponse submitChunkHierarchicalSource(HierarchicalChunkDocumentRequest request) {
-    return this.chunkOps.submitChunkHierarchicalSource(request);
-  }
-
-  @Override
   public ChunkDocumentResponse chunkSourceWithHybridChunker(HybridChunkDocumentRequest request) {
     return this.chunkOps.chunkSourceWithHybridChunker(request);
-  }
-
-  @Override
-  public TaskStatusPollResponse submitChunkHybridSource(HybridChunkDocumentRequest request) {
-    return this.chunkOps.submitChunkHybridSource(request);
   }
 
   @Override
--- a/docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/operations/AsyncOperations.java
+++ b/docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/operations/AsyncOperations.java
@@ -61,21 +61,6 @@ public abstract class AsyncOperations {
    * @param uri the endpoint URI to which the request will be sent
    * @return a {@link CompletionStage} that will be completed with the result of the asynchronous operation
    */
-  /**
-   * Submits an asynchronous task to the specified URI and returns the initial task status
-   * without polling for completion. This is useful when the caller wants to manage polling
-   * separately (e.g., a gRPC async endpoint that returns the task ID immediately).
-   *
-   * @param <I> the type of the request object being sent
-   * @param request the request object containing the data necessary to initialize the task
-   * @param uri the endpoint URI to which the request will be sent
-   * @return a {@link TaskStatusPollResponse} containing the task ID and initial status
-   */
-  protected <I> TaskStatusPollResponse submitAsync(I request, String uri) {
-    ValidationUtils.ensureNotNull(request, "request");
-    return this.httpOperations.executePost(createAsyncRequestContext(uri, request));
-  }
-
   protected <I, O> CompletionStage<O> executeAsync(I request, String uri) {
     ValidationUtils.ensureNotNull(request, "request");
 
--- a/docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/operations/ChunkOperations.java
+++ b/docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/operations/ChunkOperations.java
@@ -9,7 +9,6 @@ import ai.docling.serve.api.chunk.request.HierarchicalChunkDocumentRequest;
 import ai.docling.serve.api.chunk.request.HybridChunkDocumentRequest;
 import ai.docling.serve.api.chunk.response.ChunkDocumentResponse;
 import ai.docling.serve.api.task.request.TaskResultRequest;
-import ai.docling.serve.api.task.response.TaskStatusPollResponse;
 import ai.docling.serve.api.util.ValidationUtils;
 
 /**
@@ -42,16 +41,6 @@ public final class ChunkOperations extends AsyncOperations implements DoclingSer
   public ChunkDocumentResponse chunkSourceWithHybridChunker(HybridChunkDocumentRequest request) {
     ValidationUtils.ensureNotNull(request, "request");
     return this.httpOperations.executePost(createRequestContext("/v1/chunk/hybrid/source", request));
-  }
-
-  @Override
-  public TaskStatusPollResponse submitChunkHierarchicalSource(HierarchicalChunkDocumentRequest request) {
-    return submitAsync(request, "/v1/chunk/hierarchical/source/async");
-  }
-
-  @Override
-  public TaskStatusPollResponse submitChunkHybridSource(HybridChunkDocumentRequest request) {
-    return submitAsync(request, "/v1/chunk/hybrid/source/async");
   }
 
   @Override
--- a/docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/operations/ConvertOperations.java
+++ b/docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/operations/ConvertOperations.java
@@ -8,7 +8,6 @@ import ai.docling.serve.api.DoclingServeTaskApi;
 import ai.docling.serve.api.convert.request.ConvertDocumentRequest;
 import ai.docling.serve.api.convert.response.ConvertDocumentResponse;
 import ai.docling.serve.api.task.request.TaskResultRequest;
-import ai.docling.serve.api.task.response.TaskStatusPollResponse;
 import ai.docling.serve.api.util.ValidationUtils;
 
 /**
@@ -51,11 +50,6 @@ public final class ConvertOperations extends AsyncOperations implements DoclingS
         .responseType(ConvertDocumentResponse.class)
         .uri(uri)
         .build();
-  }
-
-  @Override
-  public TaskStatusPollResponse submitConvertSource(ConvertDocumentRequest request) {
-    return submitAsync(request, "/v1/convert/source/async");
   }
 
   @Override
--- a/docling-serve/docling-serve-grpc/build.gradle.kts
+++ b/docling-serve/docling-serve-grpc/build.gradle.kts
@@ -20,10 +20,8 @@ dependencies {
   api("com.google.protobuf:protobuf-java:$protocVersion")
   api(libs.slf4j.api)
   compileOnly("javax.annotation:javax.annotation-api:$javaxAnnotationVersion")
-  compileOnly(platform(libs.jackson.bom))
-  compileOnly(libs.jackson.annotations)
-  compileOnly(libs.jackson.databind)
-  compileOnly(libs.jackson2.databind)
+  api(platform(libs.jackson.bom))
+  api(libs.jackson.databind)
 
   testImplementation("org.mockito:mockito-core:5.17.0")
   testImplementation("org.mockito:mockito-junit-jupiter:5.17.0")
--- a/docling-serve/docling-serve-grpc/src/main/java/ai/docling/serve/grpc/v1/DoclingServeGrpcService.java
+++ b/docling-serve/docling-serve-grpc/src/main/java/ai/docling/serve/grpc/v1/DoclingServeGrpcService.java
@@ -47,7 +47,10 @@ import io.grpc.stub.StreamObserver;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.net.URI;
 import java.time.Duration;
+
+import org.jspecify.annotations.Nullable;
 
 /**
  * gRPC service implementation for DoclingServe.
@@ -68,15 +71,42 @@ public class DoclingServeGrpcService extends DoclingServeServiceGrpc.DoclingServ
   private static final Duration DEFAULT_POLL_TIMEOUT = Duration.ofMinutes(5);
 
   private final DoclingServeApi api;
+  private final @Nullable AsyncTaskSubmitter asyncSubmitter;
   private final Duration pollInterval;
   private final Duration pollTimeout;
 
   public DoclingServeGrpcService(DoclingServeApi api) {
-    this(api, DEFAULT_POLL_INTERVAL, DEFAULT_POLL_TIMEOUT);
+    this(api, (AsyncTaskSubmitter) null, DEFAULT_POLL_INTERVAL, DEFAULT_POLL_TIMEOUT);
   }
 
   public DoclingServeGrpcService(DoclingServeApi api, Duration pollInterval, Duration pollTimeout) {
+    this(api, (AsyncTaskSubmitter) null, pollInterval, pollTimeout);
+  }
+
+  public DoclingServeGrpcService(DoclingServeApi api, URI baseUrl) {
+    this(api, baseUrl, null, DEFAULT_POLL_INTERVAL, DEFAULT_POLL_TIMEOUT);
+  }
+
+  public DoclingServeGrpcService(DoclingServeApi api, URI baseUrl, @Nullable String apiKey) {
+    this(api, baseUrl, apiKey, DEFAULT_POLL_INTERVAL, DEFAULT_POLL_TIMEOUT);
+  }
+
+  public DoclingServeGrpcService(
+      DoclingServeApi api,
+      URI baseUrl,
+      @Nullable String apiKey,
+      Duration pollInterval,
+      Duration pollTimeout) {
+    this(api, new HttpAsyncTaskSubmitter(baseUrl, apiKey), pollInterval, pollTimeout);
+  }
+
+  public DoclingServeGrpcService(
+      DoclingServeApi api,
+      @Nullable AsyncTaskSubmitter asyncSubmitter,
+      Duration pollInterval,
+      Duration pollTimeout) {
     this.api = api;
+    this.asyncSubmitter = asyncSubmitter;
     this.pollInterval = pollInterval;
     this.pollTimeout = pollTimeout;
   }
@@ -122,7 +152,7 @@ public class DoclingServeGrpcService extends DoclingServeServiceGrpc.DoclingServ
       StreamObserver<ConvertSourceAsyncResponse> responseObserver) {
     try {
       var javaRequest = ServeApiMapper.toJava(request.getRequest());
-      var javaResponse = api.submitConvertSource(javaRequest);
+      var javaResponse = requireAsyncSubmitter("ConvertSourceAsync").submitConvertSource(javaRequest);
       responseObserver.onNext(ConvertSourceAsyncResponse.newBuilder()
           .setResponse(ServeApiMapper.toProto(javaResponse))
           .build());
@@ -197,7 +227,8 @@ public class DoclingServeGrpcService extends DoclingServeServiceGrpc.DoclingServ
       StreamObserver<ChunkHierarchicalSourceAsyncResponse> responseObserver) {
     try {
       var javaRequest = ServeApiMapper.toJava(request.getRequest());
-      var javaResponse = api.submitChunkHierarchicalSource(javaRequest);
+      var javaResponse = requireAsyncSubmitter("ChunkHierarchicalSourceAsync")
+          .submitChunkHierarchicalSource(javaRequest);
       responseObserver.onNext(ChunkHierarchicalSourceAsyncResponse.newBuilder()
           .setResponse(ServeApiMapper.toProto(javaResponse))
           .build());
@@ -215,7 +246,8 @@ public class DoclingServeGrpcService extends DoclingServeServiceGrpc.DoclingServ
       StreamObserver<ChunkHybridSourceAsyncResponse> responseObserver) {
     try {
       var javaRequest = ServeApiMapper.toJava(request.getRequest());
-      var javaResponse = api.submitChunkHybridSource(javaRequest);
+      var javaResponse = requireAsyncSubmitter("ChunkHybridSourceAsync")
+          .submitChunkHybridSource(javaRequest);
       responseObserver.onNext(ChunkHybridSourceAsyncResponse.newBuilder()
           .setResponse(ServeApiMapper.toProto(javaResponse))
           .build());
@@ -291,7 +323,7 @@ public class DoclingServeGrpcService extends DoclingServeServiceGrpc.DoclingServ
       StreamObserver<WatchConvertSourceResponse> responseObserver) {
     try {
       var javaRequest = ServeApiMapper.toJava(request.getRequest());
-      var initialStatus = api.submitConvertSource(javaRequest);
+      var initialStatus = requireAsyncSubmitter("WatchConvertSource").submitConvertSource(javaRequest);
       pollAndStream(initialStatus, responseObserver, "WatchConvertSource", status ->
           WatchConvertSourceResponse.newBuilder().setResponse(ServeApiMapper.toProto(status)).build());
     } catch (Exception e) {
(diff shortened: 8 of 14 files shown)


edeandrea:
Thank you @krickert for this! You've definitely put a lot of thought & time into this.

If you don't mind, let me take a few days to go through all of this. I've just returned to the office this morning after being out all week, and I want to give this the thought and review it deserves.

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeConvertApi.java
```diff
@@ -26,6 +27,19 @@ public interface DoclingServeConvertApi {
...
+   * the initial task status immediately without waiting for completion. The returned
+   * {@link TaskStatusPollResponse} contains the task ID that can be used with
+   * {@link DoclingServeTaskApi#pollTaskStatus} and {@link DoclingServeTaskApi#convertTaskResult}
+   * to track progress and retrieve results.
+   *
+   * @param request the {@link ConvertDocumentRequest} containing the source(s) and conversion options.
+   * @return a {@link TaskStatusPollResponse} containing the task ID and initial status.
+   * @throws ai.docling.serve.api.validation.ValidationException If request validation fails for any reason.
+   */
+  TaskStatusPollResponse submitConvertSource(ConvertDocumentRequest request);
```
edeandrea: I'm not sure of the purpose for this method? The existing async variant returns `CompletionStage` so that the end user doesn't have to do any of the polling/waiting. It's handled directly by the API.

Shouldn't it be the same here? I'm not sure why we want to force the user to have to poll on their own?

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeChunkApi.java
```diff
@@ -27,6 +28,17 @@ public interface DoclingServeChunkApi {
...
+  /**
+   * Submits an asynchronous hierarchical chunking task and returns the initial task status
+   * immediately without waiting for completion. The returned {@link TaskStatusPollResponse}
+   * contains the task ID that can be used to track progress and retrieve results.
+   *
+   * @param request the request containing the document source(s) and options for hierarchical chunking
+   * @return a {@link TaskStatusPollResponse} containing the task ID and initial status
+   * @throws ai.docling.serve.api.validation.ValidationException If request validation fails for any reason.
+   */
+  TaskStatusPollResponse submitChunkHierarchicalSource(HierarchicalChunkDocumentRequest request);
```
edeandrea: Similar comment as on `DoclingServeConvertApi`

I'm not sure of the purpose for this method? The existing async variant returns `CompletionStage` so that the end user doesn't have to do any of the polling/waiting. It's handled directly by the API.

Shouldn't it be the same here? I'm not sure why we want to force the user to have to poll on their own?

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeChunkApi.java
```diff
@@ -86,6 +98,18 @@ default ChunkDocumentResponse chunkFilesWithHierarchicalChunker(@Nullable Hierar
...
+   * Submits an asynchronous hybrid chunking task and returns the initial task status
+   * immediately without waiting for completion. The returned {@link TaskStatusPollResponse}
+   * contains the task ID that can be used to track progress and retrieve results.
+   *
+   * @param request the request containing the document source(s), options for conversion,
+   *                and hybrid chunking parameters
+   * @return a {@link TaskStatusPollResponse} containing the task ID and initial status
+   * @throws ai.docling.serve.api.validation.ValidationException If request validation fails for any reason.
+   */
+  TaskStatusPollResponse submitChunkHybridSource(HybridChunkDocumentRequest request);
```
edeandrea: Similar comment as above and on `DoclingServeConvertApi`

I'm not sure of the purpose for this method? The async variant returns `CompletionStage` so that the end user doesn't have to do any of the polling/waiting. It's handled directly by the API.

Shouldn't it be the same here? I'm not sure why we want to force the user to have to poll on their own?

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/DoclingServeClient.java
```diff
@@ -299,16 +299,31 @@ public ConvertDocumentResponse convertSource(ConvertDocumentRequest request) {
     return this.convertOps.convertSource(request);
   }
 
+  @Override
+  public TaskStatusPollResponse submitConvertSource(ConvertDocumentRequest request) {
+    return this.convertOps.submitConvertSource(request);
+  }
```
edeandrea: Similar comment as on `DoclingServeConvertApi` & `DoclingServeChunkApi`.

I'm not sure of the purpose for this method? The async variant returns `CompletionStage` so that the end user doesn't have to do any of the polling/waiting. It's handled directly by the API.

Shouldn't it be the same here? I'm not sure why we want to force the user to have to poll on their own?

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/DoclingServeClient.java
```diff
@@ -299,16 +299,31 @@ public ConvertDocumentResponse convertSource(ConvertDocumentRequest request) {
...
+
   @Override
   public ChunkDocumentResponse chunkSourceWithHierarchicalChunker(HierarchicalChunkDocumentRequest request) {
     return this.chunkOps.chunkSourceWithHierarchicalChunker(request);
   }
 
+  @Override
+  public TaskStatusPollResponse submitChunkHierarchicalSource(HierarchicalChunkDocumentRequest request) {
+    return this.chunkOps.submitChunkHierarchicalSource(request);
+  }
```
edeandrea: Similar comment as on `DoclingServeConvertApi` & `DoclingServeChunkApi`.

I'm not sure of the purpose for this method? The async variant returns `CompletionStage` so that the end user doesn't have to do any of the polling/waiting. It's handled directly by the API.

Shouldn't it be the same here? I'm not sure why we want to force the user to have to poll on their own?

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/DoclingServeClient.java
```diff
@@ -299,16 +299,31 @@ public ConvertDocumentResponse convertSource(ConvertDocumentRequest request) {
...
+
   @Override
   public ChunkDocumentResponse chunkSourceWithHybridChunker(HybridChunkDocumentRequest request) {
     return this.chunkOps.chunkSourceWithHybridChunker(request);
   }
 
+  @Override
+  public TaskStatusPollResponse submitChunkHybridSource(HybridChunkDocumentRequest request) {
+    return this.chunkOps.submitChunkHybridSource(request);
+  }
```
edeandrea: Similar comment as on `DoclingServeConvertApi` & `DoclingServeChunkApi`.

I'm not sure of the purpose for this method? The async variant returns `CompletionStage` so that the end user doesn't have to do any of the polling/waiting. It's handled directly by the API.

Shouldn't it be the same here? I'm not sure why we want to force the user to have to poll on their own?

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/operations/ConvertOperations.java
```diff
@@ -52,6 +53,11 @@ private <I> RequestContext<I, ConvertDocumentResponse> createRequestContext(Stri
         .build();
   }
 
+  @Override
+  public TaskStatusPollResponse submitConvertSource(ConvertDocumentRequest request) {
+    return submitAsync(request, "/v1/convert/source/async");
+  }
```
edeandrea: Similar comment as on `DoclingServeConvertApi` & `DoclingServeChunkApi`.

I'm not sure of the purpose for this method? The async variant returns `CompletionStage` so that the end user doesn't have to do any of the polling/waiting. It's handled directly by the API.

Shouldn't it be the same here? I'm not sure why we want to force the user to have to poll on their own?

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/operations/ChunkOperations.java
```diff
@@ -43,6 +44,16 @@ public ChunkDocumentResponse chunkSourceWithHybridChunker(HybridChunkDocumentReq
...
 
+  @Override
+  public TaskStatusPollResponse submitChunkHierarchicalSource(HierarchicalChunkDocumentRequest request) {
+    return submitAsync(request, "/v1/chunk/hierarchical/source/async");
+  }
+
+  @Override
+  public TaskStatusPollResponse submitChunkHybridSource(HybridChunkDocumentRequest request) {
+    return submitAsync(request, "/v1/chunk/hybrid/source/async");
+  }
```
edeandrea: Similar comment as on `DoclingServeConvertApi` & `DoclingServeChunkApi`.

I'm not sure of the purpose for this method? The async variant returns `CompletionStage` so that the end user doesn't have to do any of the polling/waiting. It's handled directly by the API.

Shouldn't it be the same here? I'm not sure why we want to force the user to have to poll on their own?

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/operations/AsyncOperations.java
```diff
@@ -61,6 +61,21 @@ protected AsyncOperations(HttpOperations httpOperations, DoclingServeTaskApi tas
...
+   *
+   * @param <I> the type of the request object being sent
+   * @param request the request object containing the data necessary to initialize the task
+   * @param uri the endpoint URI to which the request will be sent
+   * @return a {@link TaskStatusPollResponse} containing the task ID and initial status
+   */
+  protected <I> TaskStatusPollResponse submitAsync(I request, String uri) {
+    ValidationUtils.ensureNotNull(request, "request");
+    return this.httpOperations.executePost(createAsyncRequestContext(uri, request));
+  }
```
edeandrea: Similar comment as on `DoclingServeConvertApi` & `DoclingServeChunkApi`.

I'm not sure of the purpose for this method? The async variant returns `CompletionStage` so that the end user doesn't have to do any of the polling/waiting. It's handled directly by the API.

Shouldn't it be the same here? I'm not sure why we want to force the user to have to poll on their own?

docs/src/doc/docs/docling-serve/serve-grpc-client.md
```diff
@@ -0,0 +1,155 @@
...
+
+### A Quick Note on the Versioning Packaging
+
+gRPC services are defined in separate packages (e.g., `ai.docling.serve.v1`) to allow for independent versioning and evolution of the gRPC contract. This means that while the underlying `DoclingServeApi` may evolve, the gRPC service can maintain backward compatibility or introduce breaking changes in a controlled manner.
+
+This is a common practice in gRPC development.  It allows you to retain your previous records and avoid breaking changes when you upgrade the gRPC client while still maintaining a 1:1 mapping of the REST API.
+
+## Design Philosophy: REST vs. gRPC Best Practices
+
+While this module provides a functional 1:1 mapping of the Docling Serve REST API, the gRPC implementation intentionally follows Google's API Design Guide and [Buf](https://buf.build/docs/lint/overview/) linting standards.
```
edeandrea: Is there a link to Google's API Design Guide that this could be linked to?

docling-serve/docling-serve-grpc/build.gradle.kts
```diff
@@ -0,0 +1,61 @@
...
+  id("docling-lombok")
+  id("docling-release")
+  id("com.google.protobuf") version "0.9.4"
+}
+
+description = "Docling Serve gRPC API"
+
+val grpcVersion = "1.72.0"
+val protocVersion = "4.29.3"
+val javaxAnnotationVersion = "1.3.2"
```
edeandrea, line 14: We'd probably want to pull all the versions of things (& the `com.google.protobuf` plugin version) out to `libs.versions.toml` to be consistent with the rest of the project.

docling-serve/docling-serve-grpc/build.gradle.kts
```diff
@@ -0,0 +1,61 @@
...
+
+dependencies {
+  api(project(":docling-serve-api"))
+  api("io.grpc:grpc-stub:$grpcVersion")
+  api("io.grpc:grpc-protobuf:$grpcVersion")
+  api("com.google.protobuf:protobuf-java:$protocVersion")
+  api(libs.slf4j.api)
+  compileOnly("javax.annotation:javax.annotation-api:$javaxAnnotationVersion")
```
edeandrea, line 22: Shouldn't this be `jakarta.annotation:jakarta.annotation-api`?

docling-serve/docling-serve-grpc/build.gradle.kts
```diff
@@ -0,0 +1,61 @@
...
+description = "Docling Serve gRPC API"
+
+val grpcVersion = "1.72.0"
+val protocVersion = "4.29.3"
+val javaxAnnotationVersion = "1.3.2"
+
+dependencies {
+  api(project(":docling-serve-api"))
+  api("io.grpc:grpc-stub:$grpcVersion")
+  api("io.grpc:grpc-protobuf:$grpcVersion")
```
edeandrea, line 19: Probably want to define the grpc dependencies in `libs.versions.toml` to remain consistent with the rest of the project.

docling-serve/docling-serve-grpc/src/main/java/module-info.java
```diff
@@ -0,0 +1,18 @@
+module ai.docling.serve.grpc {
+  requires ai.docling.serve.api;
```
edeandrea, line 2: Should this be `requires transitive`?

docling-serve/docling-serve-grpc/src/main/java/module-info.java
```diff
@@ -0,0 +1,18 @@
...
+  requires io.grpc;
+  requires io.grpc.stub;
+  requires io.grpc.protobuf;
+  requires com.google.protobuf;
+  requires org.slf4j;
```
edeandrea, line 7: transitive?

docling-serve/docling-serve-grpc/src/main/java/module-info.java
```diff
@@ -0,0 +1,18 @@
...
+  requires java.annotation;
```
edeandrea, line 8: jakarta.annotation?

docling-serve/docling-serve-grpc/src/main/java/module-info.java
```diff
@@ -0,0 +1,18 @@
...
+
+  requires static lombok;
```
edeandrea, line 11: Lombok should not be listed (it isn't required for anything)

docling-serve/docling-serve-grpc/src/main/java/module-info.java
```diff
@@ -0,0 +1,18 @@
...
+  requires static com.fasterxml.jackson.annotation;
+  requires static com.fasterxml.jackson.databind;
+  requires static tools.jackson.databind;
```
edeandrea: Should any of the jackson packages even be listed? They're only annotations and there isn't any json marshalling in grpc, so I'm not sure why they need to be listed?

docling-serve/docling-serve-grpc/build.gradle.kts
```diff
@@ -0,0 +1,61 @@
...
+  api("com.google.protobuf:protobuf-java:$protocVersion")
+  api(libs.slf4j.api)
+  compileOnly("javax.annotation:javax.annotation-api:$javaxAnnotationVersion")
+  compileOnly(platform(libs.jackson.bom))
+  compileOnly(libs.jackson.annotations)
+  compileOnly(libs.jackson.databind)
+  compileOnly(libs.jackson2.databind)
```
edeandrea: Not sure why any of the jackson things are even needed in this module? Its only annotations which aren't used at all in grpc.

docs/src/doc/docs/docling-serve/serve-grpc-client.md
```diff
@@ -0,0 +1,155 @@
...
+
+One of the key advantages of the gRPC implementation is the "Watch" pattern. Instead of the client manually polling the status of an asynchronous task, the client can call a "Watch" RPC:
+
+- `WatchConvertSource`
+- `WatchChunkHierarchicalSource`
+- `WatchChunkHybridSource`
+
+The gRPC server will submit the task, manage the internal poll loop, and stream status updates back to the client until the task reaches a terminal state (Success or Failure).
+
+### A Quick Note on the Versioning Packaging
```
edeandrea, line 63: I'll admit I'm by no means a gRPC expert. That being said - if the grpc version depends on the underlying `DoclingServeApi` and that api changes, how can this remain backwards compatible? How would that even work, since there is a binary dependency between this module & the `docling-serve-api` module?


Or are you thinking that at some point there will be a "pure" gRPC server as part of docling itself and therefore the dependency on `DoclingServeApi` itself shouldn't even be necessary?

docling-serve/docling-serve-grpc/src/main/java/ai/docling/serve/grpc/v1/DoclingServeGrpcService.java
```diff
@@ -0,0 +1,413 @@
+package ai.docling.serve.grpc.v1;
+
+import ai.docling.serve.api.DoclingServeApi;
+import ai.docling.serve.api.chunk.response.ChunkDocumentResponse;
+import ai.docling.serve.api.convert.response.ConvertDocumentResponse;
+import ai.docling.serve.api.task.request.TaskStatusPollRequest;
+import ai.docling.serve.api.task.response.TaskStatus;
+import ai.docling.serve.api.task.response.TaskStatusPollResponse;
+import ai.docling.serve.grpc.v1.mapping.ServeApiMapper;
+import ai.docling.serve.v1.*;
```
edeandrea: No wildcard imports.

We definitely need to set up a linter in our CI, but for now there is a `.editorconfig` file in the repo. Could you please ensure that whatever IDE yoou are using adheres to the standards in there?

docling-serve/docling-serve-grpc/src/main/java/ai/docling/serve/grpc/v1/DoclingServeGrpcService.java
```diff
@@ -0,0 +1,413 @@
...
+
+  private final DoclingServeApi api;
+  private final Duration pollInterval;
+  private final Duration pollTimeout;
+
+  public DoclingServeGrpcService(DoclingServeApi api) {
+    this(api, DEFAULT_POLL_INTERVAL, DEFAULT_POLL_TIMEOUT);
+  }
+
+  public DoclingServeGrpcService(DoclingServeApi api, Duration pollInterval, Duration pollTimeout) {
```
edeandrea, line 82: The `pollInterval` & `pollTimeout` - are those for configuring the grpc "bridge"?

It might be better to follow the semantics of the other modules and offer a `Builder` for building instances of this class rather than relying on brittle constructor parameters?

docling-serve/docling-serve-grpc/src/main/java/ai/docling/serve/grpc/v1/DoclingServeGrpcService.java
```diff
@@ -0,0 +1,413 @@
...
+ * This service wraps the {@link DoclingServeApi} (REST client) and provides
+ * a gRPC interface. All RPCs are implemented as proxies:
+ * proto request → Java API → REST → Java response → proto response.
+ * <p>
+ * Includes Watch RPCs that leverage gRPC server-streaming to internally manage
+ * the poll loop, streaming each status update to the client until completion.
+ * <p>
+ * Adheres to Buf linting by using unique request/response wrappers for every RPC.
+ */
+public class DoclingServeGrpcService extends DoclingServeServiceGrpc.DoclingServeServiceImplBase {
```
edeandrea, line 67: Again I'm no gRPC expert, but the API is structured in a way where each of the different "types" of operations (chunk, convert, health, etc) are broken into smaller & separate APIs/interfaces, mostly for organization & readability purposes.

Is there a way in gRPC to structure it similarly?

docling-serve/docling-serve-grpc/src/main/java/ai/docling/serve/grpc/v1/mapping/ServeApiMapper.java
```diff
@@ -0,0 +1,719 @@
...
+/**
+ * Bidirectional mapper between proto serve types (ai.docling.serve.v1.*)
+ * and Java API models (ai.docling.serve.api.*).
+ * <p>
+ * Proto → Java: for incoming gRPC requests that need to call the REST client.
+ * Java → Proto: for REST client responses that need to be returned as gRPC responses.
+ */
+public class ServeApiMapper {
+
+  // ==================== Proto → Java (Request Mapping) ====================
```
edeandrea, line 54: Needs a `private` constructor since the class is only static methods

docling-serve/docling-serve-grpc/src/main/java/ai/docling/serve/grpc/v1/mapping/ServeApiMapper.java
```diff
@@ -0,0 +1,719 @@
...
+import java.time.Duration;
+
+/**
+ * Bidirectional mapper between proto serve types (ai.docling.serve.v1.*)
+ * and Java API models (ai.docling.serve.api.*).
+ * <p>
+ * Proto → Java: for incoming gRPC requests that need to call the REST client.
+ * Java → Proto: for REST client responses that need to be returned as gRPC responses.
+ */
+public class ServeApiMapper {
```
edeandrea, line 52: 1. Can probably be declared `final`
2. I totally understand the purpose of this class, but is there a way we can think of an easy way to maintain it rather than hard-coding everything? Maintainability is something we spent a lot of time thinking about up front. What about using something like Mapstruct? Mapstruct itself wouldn't need to be exposed outside of the module - it generates static code at build time via an annotation processor.

@ThomasVitale thoughts?

docling-serve/docling-serve-grpc/src/main/java/ai/docling/serve/grpc/v1/mapping/ServeApiMapper.java
```diff
@@ -0,0 +1,719 @@
...
+import ai.docling.serve.api.convert.response.ConvertDocumentResponse;
+import ai.docling.serve.api.convert.response.DocumentResponse;
+import ai.docling.serve.api.convert.response.ErrorItem;
+import ai.docling.serve.api.health.HealthCheckResponse;
+import ai.docling.serve.api.task.request.TaskResultRequest;
+import ai.docling.serve.api.task.request.TaskStatusPollRequest;
+import ai.docling.serve.api.task.response.TaskStatus;
+import ai.docling.serve.api.task.response.TaskStatusMetadata;
+import ai.docling.serve.api.task.response.TaskStatusPollResponse;
+import ai.docling.serve.v1.*;
```
edeandrea: Please no wildcard imports. Please try to format/adhere to the `.editorconfig`.

If you're using IntelliJ you can configure it to automatically respect it.

docling-serve/docling-serve-grpc/src/main/java/ai/docling/serve/grpc/v1/mapping/DoclingDocumentMapper.java
```diff
@@ -0,0 +1,875 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Maps ai.docling.core.DoclingDocument (Java) to ai.docling.core.v1.DoclingDocument (Proto).
+ * <p>
+ * This is a proper field-by-field mapper with NO JSON.
+ * Each field is explicitly mapped to maintain strong typing throughout the gRPC stack.
+ */
+@SuppressWarnings("DataFlowIssue")
+public class DoclingDocumentMapper {
```
edeandrea, line 56: Pretty much all the same comments I left in `ServeApiMapper`

docling-serve/docling-serve-grpc/src/main/java/ai/docling/serve/grpc/v1/mapping/ProtoMapping.java
```diff
@@ -0,0 +1,27 @@
+package ai.docling.serve.grpc.v1.mapping;
+
+import java.util.function.Consumer;
+
+/**
+ * Shared utilities for proto ↔ Java mapping.
+ */
+final class ProtoMapping {
```
edeandrea, line 8: There's already a `Utils` class in `docling-serve-api` that you could add this to.

edeandrea (commented):
Thank you very much @krickert for this! You've definitely put a lot of thought & work into this, so its much appreciated.

Sorry for the delay in my review. I wanted to look over it, give it some time to settle in my brain and think about it a bit.

I'd also like @ThomasVitale to look through this as well.

There are a few small general comments I have that I saw in lots of places (note to self - I need to update the contributing guidelines with these so that people are aware):
1. No wildcard imports please
2. While we do need to put some linting in place, there is a `.editorconfig` file in the repository. It would be great if you could configure your IDE to adhere to whats in there.

The biggest part of this that I want to make sure I understand is that `DoclingServeGrpcService` is a gRPC server that wraps `DoclingServeApi`. Then the gRPC definitions from the protobuf files can interact with the server?

The thought being that at some point if docling itself supported a gRPC server out of the box, then `DoclingServeGrpcService` (& the dependency on `docling-serve-api`) wouldn't even be needed?

If thats the case, should we structure this similar to how we've structured the REST side? One module just for the API (which in the case of gRPC would mostly just be the protobuf definitions)? And then a client module, which would be the wrapper thing thats here?

That way, when/if docling has a grpc server out of the box, the client module would just get deprecated and cease to exist?

@ThomasVitale I'm curious as to your overall thoughts as well.

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/operations/AsyncOperations.java
krickert: I will delete it.. it's not needed.  I added both figuring we would choose one - and added the one where we did polling after the fact.  

Sidenote: when I tried similar grpc conversions on 2 other (very popular) OSS projects, the maintainers for both were 100% convinced gRPC:Rest should be 1:1.  So I'm happy to see that a Quarkus committer wouldn't share that assessment.

docling-serve/docling-serve-grpc/src/main/java/ai/docling/serve/grpc/v1/mapping/ProtoMapping.java
krickert: Will do...

docling-serve/docling-serve-grpc/src/main/java/module-info.java
krickert: Apologies on this one - it was auto generated by the IDE.  I'll clean it up.

docling-serve/docling-serve-grpc/build.gradle.kts
krickert: No problem

docling-serve/docling-serve-grpc/build.gradle.kts
krickert: Yes. Old habits, some of the older grpc stuck with javax.annotation until semi-recently. I think you're right.

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeConvertApi.java
krickert: Done

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeChunkApi.java
krickert: Done

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/DoclingServeClient.java
krickert: done

docling-serve/docling-serve-grpc/src/main/java/ai/docling/serve/grpc/v1/mapping/ServeApiMapper.java
krickert: Done

docling-serve/docling-serve-grpc/src/main/java/ai/docling/serve/grpc/v1/mapping/ServeApiMapper.java
krickert: `DoclingServeGrpcService`. Yup - a bridge/wrapper. I actually have an open ticket with Docling to support gRPC natively (they haven't bit yet), so the goal of this implementation is to serve as a proof-of-concept to motivate them to move in that direction.

To your question about structure: While mirroring the REST setup feels intuitively consistent, I strictly avoid a 1:1 mapping between REST and Protobuf. They play by different rules, and treating them the same introduces long-term risks.

Here is why I want to stick to the standard "v1" Protobuf definition approach:

Best to think of protobufs as Data, not just APIs. Unlike REST, which is ephemeral, Protobuf messages are often used as immutable records for storage (Kafka, logs, cold archival). I plan to use these definitions for long-term storage of Docling outputs.  Makes sense too right?  It's a document service type output.  They even want to standardize this model more, which suggests it'll settle down and not change too much anyway.

If we couple the `.proto` package structure to our current temporary wrapper structure (to match REST), we bake implementation details into the data.

Therefore --- If Docling later releases a native server and we switch to it, or if we refactor our wrapper, we break backward compatibility with every record we’ve ever stored (and would've had to make definition changes anyway). I want to avoid needing a complex migration/mapping class just to read data we saved a year ago.

I also avoid "REST-in-Protobuf" syntax. When we try to mirror REST 1:1, we usually end up with "bad" Protobuf definitions. We miss out on streaming capabilities, we end up with unnecessary empty request/response objects, or we create definitions that mimic a specific language's (Java/Python) object structure rather than a language-neutral contract.

The API module should contain pure, immutable `v1` definitions that look like they came from a native Docling server, agnostic of our specific wrapper.

If/when Docling releases a server, our wrapper module (`DoclingServeGrpcService`) effectively becomes obsolete and can be deleted. However, because we kept the **definitions** standard and separate:

* We won't have to change the data format.
* The generated clients (stubs) will continue to work with zero coding changes.
* We avoid the old SOAP-style integration nightmares; the auto-generated stubs just handle the switch.

So if you're ok with this - let’s keep the definitions (the Schema) distinct from the implementation (the Wrapper). The Schema should be designed as if the "perfect" Docling gRPC server already exists and not let the JSON/REST bleed into it. This allows us to use these objects in Kafka, mobile apps, or storage without worrying about the internal structure of our temporary bridge service.

I know it's not as intuitive, but you really don't even need to make separate tags and releases of protobufs if you design it well since the definitions should always line up and a schema registry can automatically track the changes.

docling-serve/docling-serve-grpc/src/main/java/ai/docling/serve/grpc/v1/mapping/ServeApiMapper.java
krickert: I'll setup intellj to lint better..

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/operations/AsyncOperations.java
edeandrea: I think that each technology has its purposes, and there are also other considerations at play, some of which are non-technical.

For me personally, I feel that we should look for whatever the best solution is, while still adhereing to what we consider our "guiding principles" for this project.

I'm always open to new ideas on how to do things - I'm certainly not an expert in everything, so if someone else has a good idea I'm always open to considering it.

That being said - we'll see how this whole thing shakes out :)

docling-serve/docling-serve-grpc/build.gradle.kts
edeandrea: On 2nd thought, maybe they are needed, since the gRPC server is doing rest and you'll need jackson. But Jackson shouldn't need to be exposed outside of this module (its an implementation detail).

krickert:
> > Still workin' on the changes - I can get to the rest tonight. Worked out the obvious ones already, looked into [MapStruct](https://github.com/entur/mapstruct-spi-protobuf) protobuf plugin, it would be adding another layer though and it's just one API we're mirroring. If the plan is to mirror many, I'd recommend this but the scope of the mapping was 3 classes. I'm going to run this through potentially 100s of millions of docs, so any bugs are going to be found. Finally, it would be fastest this way and any "gotchas" that come out of the REST api can be mapped here. But maintaining the API isn't that bad - it took me less than a day to write up the API.
> 
> I appreciate you looking into things!
> 
> Re: the mapping stuff - there is a LOT of logic in the mapping classes, which looks like it is mapping between the protobuf objects and the `DoclingServeApi` objects. It is "brittle" code that would most likely be dependent on the protobuf version.
> 
> @ThomasVitale and I are only 2 people working part time on this project, so at the beginning we said that we really needed to favor maintainability as one of the top concerns.
> 
> I wasn't aware there was a mapstruct protobuf thing, I was merely talking about [Mapstruct](https://mapstruct.org/) the java mapper. It provides a nice way to map from one object structure to another in a declarative fashion.
> 
> Now I'm not saying i want to force it, but just something to look into.

I think it's worth looking into.  I'm looking for a better way too - I use CEL on protobufs on my mainline project to handle my mappings between protobufs... it's faster than MapStruct I'm sure - but would be worth looking into as well.  CEL works on both protobufs and pojos, so I think that would be a bit less brittle because it's an industry standard and not just a java library - so I think it can map between the REST Jackson objects to the Protobufs well and even put expressions (like in my case, the default values).

Another suggestion - if you consider it - having the grpc layer be the client instead of a java layer.   I think this layer is still needed though, until we get a true grpc client from the server then I'd go that route instead.   But even with this, you can just use the grpc client then you don't have to maintain any code just the mapping layer and offer all the functionality of the server on any client... it'll also make it attractive to those who have a grpc environment and would prefer this layer over rest.  just a thought :)

krickert:
Closing, working on https://github.com/ai-pipestream/docling-serve/tree/grpc-native-converter in the grpc layer.  When it is ready, I can open a PR for comment there.

PR #366, opened 2026-02-25
ci: set up Spotless linting in CI (#351)

### LINKED ISSUE #351: Set up code linting as part of DevOps processes
Currently we don't do any linting when running CI jobs. We'd like to do that.

This issue is about figuring out what the correct tool is (CheckStyle/Spotless/etc) and integrating it into the CI pipeline.

KaleabArgaw:
Fixes #351

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -124,6 +124,8 @@ jobs:
     if: github.event_name == 'pull_request'
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
 
       - name: Reclaim disk space
         working-directory: .github/scripts


.github/workflows/build.yml
```diff
@@ -18,6 +18,38 @@ defaults:
     shell: bash
 
 jobs:
+  lint:
```
edeandrea: I think it would be better for spotless to have its own workflow file. Most likely it will only run on pull requests

.github/workflows/build.yml
```diff
@@ -18,6 +18,38 @@ defaults:
...
+      - name: spotless-apply
+        run: ./gradlew --no-daemon -Pjava.version=17 spotlessApply
+
+      - name: Verify formatting is clean
+        run: |
+          git diff --exit-code || {
+            echo "Formatting issues were fixed by Spotless. Run './gradlew --no-daemon spotlessApply' locally and commit the result."
+            exit 1
+          }
+
```
edeandrea, line 120: I'm not sure we want to have spotless "fix" things on CI. Instead I think finding issues and throwing an error state will be better.

My intent is the check will become part of the ruleset that must pass in order for a PR to be merged.

If it does fail, displaying something to the user on how they can fix it in a subsequent commit would be great!

buildSrc/src/main/kotlin/docling-java-shared.gradle.kts
```diff
@@ -41,6 +42,19 @@ jacoco {
   toolVersion = libs.findVersion("jacoco").get().toString()
 }
 
+spotless {
+  // Roll out linting incrementally by checking only files changed from main.
+  ratchetFrom("origin/main")
```
edeandrea, line 46: This is causing a build failure in the `docs` module that needs to be investigated.

buildSrc/src/main/kotlin/docling-java-shared.gradle.kts
```diff
@@ -41,6 +42,19 @@ jacoco {
...
+
+  java {
+    target("src/*/java/**/*.java")
+    removeUnusedImports()
+    formatAnnotations()
+    trimTrailingWhitespace()
+    endWithNewline()
+  }
+}
```
edeandrea, line 57: I haven't read up on spotless and its configuration - is there a way to specify the configuration outside of the gradle dsl? And is there a way to get it to respect what we have in `.editorconfig`?

Ideally I'd prefer control over formatting/etc to be defined in a single place, whether thats `.editorconfig` or somewhere else. It would also be nice if when a user loads the project into an IDE, the formatting rules are respected by the IDE. Anything we can do to make that easier is a win.

KaleabArgaw: Yeah we can govern it using `.editorconfig`

.github/workflows/build.yml
KaleabArgaw: I have moved out spotless to a separate file and it will only now check on PR @edeandrea

edeandrea: I see the separate `spotless.yml`, so I'm not sure why there are any changes to this file then?

The diff is showing that the whole file has changed?

CONTRIBUTING.md
edeandrea, line 1: Its showing that the entire file has changed?

buildSrc/src/main/kotlin/docling-java-shared.gradle.kts
edeandrea: Does it use `.editorconfig` by default if you don't add any additional config for java?

I see [in the documentation](https://github.com/diffplug/spotless/blob/main/plugin-gradle/README.md#ktlint) it works that way for kotlin (ktlint)

gradle/libs.versions.toml
```diff
@@ -9,10 +9,11 @@ junit = "6.0.3"
 lombok = "1.18.42"
 lombok-gradle = "9.2.0"
 module-info = "1.14"
+spotless = "7.2.1"
```
edeandrea: Should we be using the [latest version](https://plugins.gradle.org/plugin/com.diffplug.gradle.spotless) (8.2.1)?

KaleabArgaw: Yeah i think using the latest is better will make this change.

CONTRIBUTING.md
KaleabArgaw: It was a merge issue you can check it now @edeandrea and thanks.

PR #511, opened 2026-05-22
build: Run tests on Java 26

ThomasVitale:


edeandrea:
Should we really be building against Java 26? We probably only want to support LTS version...

ThomasVitale:
@edeandrea that would be a good thing to decide and document. Don't we wanna build against LTS versions + the current version of Java to make sure it works? I started a discussion in https://github.com/docling-project/docling-java/discussions/531 so we can talk about that.

PR #535, opened 2026-06-12
feat(api): add PresignedUrlTarget for v1.22 S3 delivery

dan-m62:
`docling-serve` v1.22.0+ introduces a server-managed presigned-URL delivery mode: the operator configures `DOCLING_SERVE_ARTIFACT_STORAGE_*` settings, clients request `{"target": {"kind": "presigned_url"}}`, and the server uploads each output artifact to its own bucket and returns time-limited presigned GET URLs in the response.

The response side of this surface (`PreSignedUrlConvertDocumentResponse`) exists via #388 but the request side was not added at the same time. Because `Target` is sealed, downstream users cannot supply their own subtype — they either bypass the SDK to hand-build the request JSON, or fork.

This commit adds `PresignedUrlTarget` as a permitted subtype of Target, mirroring the existing trivial-marker pattern used by `InBodyTarget` and `ZipTarget` i.e.:

  - PresignedUrlTarget.java: empty marker with the standard builder
  - Target.java: extend permits (alphabetical), @JsonSubTypes, and Kind enum with the `presigned_url` discriminator
  - PresignedUrlTargetTests.java: builder + Target-subtype + toBuilder roundtrip assertions, matching the test scope of the other target tests in this module


edeandrea:
Hi @dan-m62 thank you for this!

Something I noticed - the `PresignedUrlTargetTests` that you added - they don't really test that the marshalling nor that the requests/responses actually work. 

`whenBuiltThenIsTargetSubtype` and `whenBuiltThenInstanceCreated` tests are pretty pointless to be honest. It would be better if there were some functional tests in [`AbstractDoclingServeClientTests`](https://github.com/docling-project/docling-java/blob/main/docling-serve/docling-serve-client/src/test/java/ai/docling/serve/client/AbstractDoclingServeClientTests.java) that verified that the docling-serve instance actually did things the way it was supposed to based on the input.

edeandrea:
This is 1/2 the implementation of #541

edeandrea:
@all-contributors add @dan-m62 for code, idea

edeandrea:
Closing in favor of #543, which builds on the foundation you laid here. Thank you @dan-m62 for your work on `PresignedUrlTarget` — it provided the starting point for the consolidated PR.

PR #537, opened 2026-06-14
feat(api): add PreSignedUrlConvertResponse for PresignedUrlTarget

dan-m62:
Companion to #535, completing the request/response round-trip for the `PresignedUrlTarget` flow introduced in `docling-serve` v1.22.0+. When a client submits `{"target": {"kind": "presigned_url"}}`, the
  server's response shape (`PresignedUrlConvertResponse` on the Python side) carries:

  - aggregate counts (`num_converted` / `num_succeeded` / `num_partially_succeeded` / `num_failed` / `processing_time`)
  - a `documents[]` array with per-source artifact metadata: `source_index`, `source_uri`, `filename`, `status`, `errors`, `timings`, and `artifacts[]` (per output format: `artifact_type`, `mime_type`, `uri`,
  `url_expires_at`)

  This is structurally distinct from the existing `PreSignedUrlConvertDocumentResponse` (added by #388), which carries only the aggregate counts and is what the server returns for `s3` / `put` targets. The Python
   equivalents are similarly two distinct classes — `PresignedUrlConvertDocumentResponse` (counts only) and `PresignedUrlConvertResponse` (counts + documents). The existing Java class is unchanged.

  This commit adds:

  - `PreSignedUrlConvertResponse.java`: new sibling under the sealed `ConvertDocumentResponse` hierarchy, fingerprinted by the presence of `documents` (Jackson `JsonTypeInfo.Id.DEDUCTION`, consistent with the
  rest of this hierarchy — the wire response carries no `kind` discriminator)
  - `DocumentArtifactItem.java`, `ArtifactRef.java`, `ArtifactType.java`, `ConversionStatus.java`, `ProfilingItem.java`, `ProfilingScope.java`: supporting types, fields verified 1:1 against `docling-slim` 2.99.0
  - `ResponseType.java`: extend with `PRE_SIGNED_URL_RESPONSE`
  - `ConvertDocumentResponse.java`: extend `permits` and `@JsonSubTypes`
  - `ArtifactRefTests`, `DocumentArtifactItemTests`, and two new methods on `ConvertDocumentResponseTests` covering builder + null-collection + immutability assertions, matching the test scope of the other
  response tests in this module


edeandrea:
Hi @dan-m62 thank you for this work! I have similar comments here about the tests [as I did in #535](https://github.com/docling-project/docling-java/pull/535#issuecomment-4702476128). The tests here really don't provide any indication that the json marshalling is correct and that it serializes/deserializes properly.

t would be better if there were some functional tests in [AbstractDoclingServeClientTests](https://github.com/docling-project/docling-java/blob/main/docling-serve/docling-serve-client/src/test/java/ai/docling/serve/client/AbstractDoclingServeClientTests.java) that verified that the docling-serve instance actually did things the way it was supposed to based on the input.

edeandrea:
This is the other half of #541

edeandrea:
Closing in favor of #543, which builds on the foundation you laid here. Thank you @dan-m62 for your work on `PreSignedUrlConvertResponse` and the supporting model types — it provided the starting point for the consolidated PR.

Issue #350, opened 2026-02-24
Increase code coverage


edeandrea:
We'd like to have more tests to increase our code coverage. Also more tests around other use cases as well (other parts of the APIs).

lemosdevv:
hello, I would like to help

codingkiddo:
Hi @edeandrea, I would like to work on #350 by adding a small unit test around `DoclingServeClientBuilderFactory`.

The test will cover the edge case where neither Jackson 2 nor Jackson 3 is visible to the supplied classloader, and will assert that the factory throws a helpful `IllegalStateException` with dependency guidance.

Does this sound okay?

edeandrea:
> Hi [@edeandrea](https://github.com/edeandrea), I would like to work on [#350](https://github.com/docling-project/docling-java/issues/350) by adding a small unit test around `DoclingServeClientBuilderFactory`.
> 
> The test will cover the edge case where neither Jackson 2 nor Jackson 3 is visible to the supplied classloader, and will assert that the factory throws a helpful `IllegalStateException` with dependency guidance.
> 
> Does this sound okay?

Sounds great! Thank you!

PR #570, opened 2026-06-27
test(version-tests): add tests for client, domain, and service layers

margi212:
## Summary

Adds comprehensive unit test coverage across the `client`, `domain`, and `service` layers of the `docling-version-tests` module.

---

### New Test Classes

| File | Class Under Test | # Tests |
|------|-----------------|---------|
| `RegistryClientFactoryTests` | `RegistryClientFactory` | 8 |
| `GHCRClientLoggerTests` | `GHCRClientLogger` | 16 |
| `TagsTestRequestTests` | `TagsTestRequest` | 17 |
| `NamedThreadFactoryTests` | `NamedThreadFactory` | 7 |
| `WorkParallelizerTests` | `WorkParallelizer` | 10 |

### Extended Test Classes

| File | What was added |
|------|---------------|
| `TagTestResultTests` | 9 new tests for `TagsTestResults` sorting, `from()` factory, `toBuilder()`, `setResults()`/`clearResults()`, failure detection, and nested/long stack trace handling |

---

## Coverage Highlights

- **`RegistryClientFactory`** — validates correct client dispatch for `ghcr.io`, plus rejection of unsupported, null, empty, whitespace-padded, and case-mismatched registry strings; verifies client instance reuse.
- **`GHCRClientLogger`** — validates conditional request/response logging, masking of `token` fields in JSON bodies, `Authorization` and `Set-Cookie` header redaction, and graceful handling of null, empty, non-JSON, and JSON-array bodies.
- **`TagsTestRequest`** — validates builder construction, `toBuilder()` round-trip fidelity, method chaining, override behavior, and mandatory `executor` field enforcement.
- **`NamedThreadFactory`** — validates thread naming with incrementing suffix, isolated counters per factory instance, and thread-safe concurrent creation.
- **`WorkParallelizer`** — validates parallel `transform` and `run` semantics across empty, single, multi-item, and large (100-item) workloads with type-safe generic transformation.
- **`TagTestResult` / `TagsTestResults`** — extends existing tests with version-sorted result sets, factory construction from a `TagsTestRequest`, mutable builder operations, and failure presence detection.

---

## Test Design Notes

- All tests follow **Arrange / Act / Assert** structure with `assertThat` (AssertJ).
- Edge cases (null, empty, whitespace) are tested explicitly to guard against NPEs and silent failures.
- Concurrent tests use `CountDownLatch` / `AtomicInteger` to avoid race conditions in assertions.
- No production code was modified; this PR is purely additive test coverage.

Related to #350


docling-testing/docling-version-tests/src/test/java/ai/docling/client/tester/service/WorkParallelizerTests.java
```diff
@@ -0,0 +1,189 @@
...
+
+        assertThat(results)
+                .hasSize(1)
+                .containsExactly("TEST");
+
+        executor.shutdown();
+    }
+
+    @Test
+    void shouldRunItemsInParallel() {
```
edeandrea, line 81: I'm not sure what the purpose of this test is or what its trying to prove?

docling-testing/docling-version-tests/src/test/java/ai/docling/client/tester/service/WorkParallelizerTests.java
```diff
@@ -0,0 +1,189 @@
...
+        assertThat(processedItems)
+                .hasSize(5)
+                .containsExactlyInAnyOrder(1, 2, 3, 4, 5);
+
+        executor.shutdown();
+    }
+
+    @Test
+    void shouldRunEmptyList() {
+        var executor = Executors.newSingleThreadExecutor();
```
edeandrea: I'm not sure what the purpose of this test is or what its trying to prove?

docling-testing/docling-version-tests/src/test/java/ai/docling/client/tester/service/WorkParallelizerTests.java
```diff
@@ -0,0 +1,189 @@
...
+                items,
+                item -> counter.incrementAndGet());
+
+        assertThat(counter.get()).isZero();
+
+        executor.shutdown();
+    }
+
+    @Test
+    void shouldRunSingleItem() {
```
edeandrea: I'm not sure what the purpose of this test is or what its trying to prove?

docling-testing/docling-version-tests/src/test/java/ai/docling/client/tester/service/WorkParallelizerTests.java
```diff
@@ -0,0 +1,189 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.junit.jupiter.api.Test;
+
+class WorkParallelizerTests {
```
edeandrea, line 18: I feel like there are many overlapping tests here that aren't needed and don't really prove anything. The first two which do something on a non-empty & empty list are probably enough. The rest of them are really different variations of the same test.

docling-testing/docling-version-tests/src/test/java/ai/docling/client/tester/service/NamedThreadFactoryTests.java
```diff
@@ -0,0 +1,106 @@
+package ai.docling.client.tester.service;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+
+class NamedThreadFactoryTests {
```
edeandrea, line 8: I feel like there are many overlapping tests here that aren't needed and don't really prove anything. The first one which creates a named thread and verifies the thread name is probably sufficient. The rest of them are really different variations of the same test.

docling-testing/docling-version-tests/src/test/java/ai/docling/client/tester/service/NamedThreadFactoryTests.java
```diff
@@ -0,0 +1,106 @@
...
+    @Test
+    void shouldCreateThreadWithName() {
+        var factory = new NamedThreadFactory("test-thread");
+        Runnable runnable = () -> {
+        };
+
+        Thread thread = factory.newThread(runnable);
+
+        assertThat(thread).isNotNull();
+        assertThat(thread.getName()).isEqualTo("test-thread-1");
```
edeandrea: The assertions could be written more fluently.

```java
assertThat(thread)
    .isNotNull()
    .extracting(Thread::getName)
    .isEqualTo("test-thread-1")
```

It would probably be better to do a `startsWith("test-thread")` assertion rather than an `isEqualTo("test-thread-1")` because the `isEqualTo` assumes some internal implementations as to the way threads are named/numbered. If the jdk changes that in future versions then these tests will break. All we care about is that the thread starts with a given name.

docling-testing/docling-version-tests/src/test/java/ai/docling/client/tester/client/ghcr/GHCRClientLoggerTests.java
```diff
@@ -0,0 +1,373 @@
...
+
+    @BeforeEach
+    void setUp() {
+        config = mock(Config.class);
+        logger = new GHCRClientLogger(config);
+    }
+
+    @Test
+    void shouldSetBodySize() {
+        logger.setBodySize(1024);
```
edeandrea: This test doesn't have any assertions. What is it proving?

docling-testing/docling-version-tests/src/test/java/ai/docling/client/tester/client/ghcr/GHCRClientLoggerTests.java
```diff
@@ -0,0 +1,373 @@
...
+import ai.docling.client.tester.config.Config;
+
+import io.vertx.core.Handler;
+import io.vertx.core.MultiMap;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.http.HttpClientRequest;
+import io.vertx.core.http.HttpClientResponse;
+import io.vertx.core.http.HttpMethod;
+
+class GHCRClientLoggerTests {
```
edeandrea, line 26: There are a lot of repetitive tests here that are variations on the same test (response status, headers, etc).

docling-testing/docling-version-tests/src/test/java/ai/docling/client/tester/client/RegistryClientFactoryTests.java
```diff
@@ -0,0 +1,82 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.Mockito.mock;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import ai.docling.client.tester.client.ghcr.GHCRClient;
+
+class RegistryClientFactoryTests {
```
edeandrea, line 15: I'd say the tests are all valid, but many of them are the same test just with different inputs and the same expectations.

They could be cleaned up using JUnit's `ParameterizedTest` - that way you only have to write the test logic once and pass different inputs into the test.

docling-testing/docling-version-tests/src/test/java/ai/docling/client/tester/domain/TagsTestRequestTests.java
```diff
@@ -0,0 +1,249 @@
...
+                .cleanupContainerImages(true)
+                .tags(List.of("v1.0.0", "v1.1.0"))
+                .build();
+
+        assertThat(request).isNotNull();
+        assertThat(request.registry()).isEqualTo("ghcr.io");
+        assertThat(request.image()).isEqualTo("docling-project/docling-serve");
+        assertThat(request.executor()).isEqualTo(TEST_EXECUTOR);
+        assertThat(request.cleanupContainerImages()).isTrue();
+        assertThat(request.tags()).containsExactly("v1.0.0", "v1.1.0");
```
edeandrea: Rather than a handful of individual asserts like this you could do something like this instead:

```java
var request = ... // Build the request using the builder as you have
var expected = new TagsTestRequest("ghcr.io", "docling-project/docling-serve", TEST_EXECUTOR, true, List.of("v.1.0.0", "v1.1.0"));

assertThat(request)
    .isNotNull()
    .usingRecursiveComparison()
    .isEqualTo(expected);
```

AssertJ provides a nice way to perform equality by doing field-by-field comparisons (see https://assertj.github.io/doc/#assertj-core-recursive-comparison).

docling-testing/docling-version-tests/src/test/java/ai/docling/client/tester/domain/TagsTestRequestTests.java
```diff
@@ -0,0 +1,249 @@
...
+                .tags(List.of("v1.0.0"))
+                .build();
+
+        var modifiedRequest = originalRequest.toBuilder()
+                .registry("quay.io")
+                .cleanupContainerImages(true)
+                .tags(List.of("v2.0.0"))
+                .build();
+
+        assertThat(originalRequest.registry()).isEqualTo("ghcr.io");
```
edeandrea: Similar comment as above regarding the usage of recursive comparisons

docling-testing/docling-version-tests/src/test/java/ai/docling/client/tester/domain/TagsTestRequestTests.java
```diff
@@ -0,0 +1,249 @@
...
+
+        var copy = original.toBuilder().build();
+
+        assertThat(copy.registry()).isEqualTo(original.registry());
+        assertThat(copy.image()).isEqualTo(original.image());
+        assertThat(copy.executor()).isEqualTo(original.executor());
+        assertThat(copy.cleanupContainerImages()).isEqualTo(original.cleanupContainerImages());
+        assertThat(copy.tags()).isEqualTo(original.tags());
+    }
+
```
edeandrea: The rest of these tests are just variants of the same tests and aren't needed.

edeandrea (changes requested):
Hi @margi212 Thanks for doing this! I've made some specific comments in certain places but I'll add some general comments here.

You may be new to AssertJ and thats totally fine, but if possible could you try to keep the assertions more fluent? Thats one of the main powers of AssertJ - rather than having lots of single `assertThat` statements, assertions can be combined in a fluent style.

The other thing I noticed was that there were lots of tests that were just variations on the same test. Many of them are unneeded, and others could be re-written using parameterized tests (same test assertion(s) just with different inputs).

edeandrea (dismissed):
Thanks for this @margi212 !

edeandrea:
@margi212 could you please fix the test failures? The Semantic PR check is ok - I can fix that when I merge, but the other checks are failing.

margi212:
> @margi212 could you please fix the test failures? The Semantic PR check is ok - I can fix that when I merge, but the other checks are failing.

@edeandrea , can you please check once again?

edeandrea (approved):
Thanks @margi212 !

edeandrea:
@all-contributors add @margi212 for test

PR #691, opened 2026-09-22
feat(serve-client): support custom Executor for async operations

### LINKED ISSUE #664: Add support for custom Executor
Currently all the blocking API methods run on the calling thread, and the async methods run on the fork-join pool. We should allow for passing in (or building the client) with a custom `Executor` for async operations.

Naman-Gururani:
Closes #664

Implements the design discussed in the issue.

### What changed

- `DoclingApiBuilder.asyncExecutor(Executor)`: new builder method, implemented by `DoclingServeClientBuilder`. It's a `default` method throwing `UnsupportedOperationException`, so existing third-party builder implementations keep compiling. `null` is rejected with `IllegalArgumentException`, like the other builder arguments.
- The executor stays `@Nullable` down to `AsyncOperations`, which uses it for every step of an async operation: submitting the task, each status poll, the delayed re-poll and retrieving the result. When no executor is set it uses the 1-arg `supplyAsync` / 2-arg `delayedExecutor`, so the behaviour is exactly the previous one (`CompletableFuture`'s default executor is not always `ForkJoinPool.commonPool()`).
- `toBuilder()` carries the executor over. The client never shuts it down – its lifecycle belongs to the caller.
- `ConvertOperations`, `ChunkOperations` and `AsyncOperations` gain constructor overloads taking the executor; their existing constructors are unchanged.
- Javadoc (including a note on direct executors) and a "What's New" entry.

### Tests

- `AbstractDoclingServeClientAsyncExecutorTests`, run for both clients by `DoclingServeJackson2ClientAsyncExecutorTests` / `DoclingServeJackson3ClientAsyncExecutorTests` (WireMock), drives tasks through submit → `started` poll → delayed re-poll → final status (→ result) and counts the tasks submitted to the executor: 5 for successful convert and hybrid-chunk tasks, 4 for a failed task. It also covers the default executor, `toBuilder()` and `null`.
- `DoclingServeApiTests.asyncExecutorIsUnsupportedByDefault` covers the default method.

`./gradlew spotlessCheck` and the new tests pass locally.

Note: `spotlessApply` re-formatted a few pre-existing lines in the touched files (`DoclingServeApi`, `DoclingServeClient`, `AsyncOperations`, `ConvertOperations`), which is why the diff is a bit larger than the change itself. Happy to squash the review commits once this is approved.


docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/operations/AsyncOperations.java
```diff
@@ -27,12 +30,19 @@ public abstract class AsyncOperations {
   private final DoclingServeTaskApi taskApi;
   private final Duration asyncPollInterval;
   private final Duration asyncTimeout;
+  private final Executor asyncExecutor;
 
   protected AsyncOperations(HttpOperations httpOperations, DoclingServeTaskApi taskApi, Duration asyncPollInterval, Duration asyncTimeout) {
+    this(httpOperations, taskApi, asyncPollInterval, asyncTimeout, null);
+  }
+
+  protected AsyncOperations(HttpOperations httpOperations, DoclingServeTaskApi taskApi, Duration asyncPollInterval, Duration asyncTimeout, @Nullable Executor asyncExecutor) {
```
edeandrea: I think it would be better to guarantee that the `Executor` is not nullable. The null check should happen before this constructor.

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/DoclingServeClient.java
```diff
@@ -90,6 +91,7 @@ public abstract class DoclingServeClient extends HttpOperations implements Docli
   private final Duration readTimeout;
   private final Duration asyncPollInterval;
   private final Duration asyncTimeout;
+  private final @Nullable Executor asyncExecutor;
```
edeandrea: This shouldn't be `@Nullable`

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/DoclingServeClient.java
```diff
@@ -129,12 +131,13 @@ protected DoclingServeClient(DoclingServeClientBuilder builder) {
     this.apiKey = builder.apiKey;
     this.asyncPollInterval = builder.asyncPollInterval;
     this.asyncTimeout = builder.asyncTimeout;
+    this.asyncExecutor = builder.asyncExecutor;
```
edeandrea: I would do the `null` check here and set the default, that way its always set.

edeandrea (commented):
Thanks @Naman-Gururani for the PR! I think it would be better for the `Executor` to never be `@Nullable`, except in the builder itself. The `null` check and setting the default should happen in the `DoclingServeClient` constructor I think.

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/operations/AsyncOperations.java
edeandrea: forget this - sorry

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/DoclingServeClient.java
edeandrea: forget this - sorry

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/DoclingServeClient.java
edeandrea: forget this - sorry

Naman-Gururani:
Thanks @edeandrea, done in 90321a4: the executor is now only `@Nullable` in the builder, the `DoclingServeClient` constructor does the null check and sets the `ForkJoinPool.commonPool()` default, and `AsyncOperations` / `ConvertOperations` / `ChunkOperations` take a required `Executor` (the old constructors are replaced since `DoclingServeClient` is their only caller). I'll squash before merge if you'd like.

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/operations/AsyncOperations.java
```diff
@@ -129,11 +131,10 @@ private <O> CompletionStage<O> pollTaskStatus(TaskStatusPollResponse statusRespo
...
 
       default ->
-        // Still in progress (PENDING or STARTED), schedule next poll after delay
-        CompletableFuture.supplyAsync(
-            () -> null,
-            CompletableFuture.delayedExecutor(this.asyncPollInterval.toMillis(), TimeUnit.MILLISECONDS)
-        ).thenCompose(v -> pollTaskUntilComplete(statusResponse, startTime));
+          // Still in progress (PENDING or STARTED), schedule next poll after delay
+          CompletableFuture.supplyAsync(
+              () -> null, CompletableFuture.delayedExecutor(this.asyncPollInterval.toMillis(), TimeUnit.MILLISECONDS, this.asyncExecutor)
```
edeandrea: Please don't materialize a default executor here (or in `DoclingServeClient`). Keep the field `@Nullable Executor` and branch at each use site, passing it to `supplyAsync` only when it was actually set:

```java
private final @Nullable Executor asyncExecutor;

private <T> CompletableFuture<T> supply(Supplier<T> supplier) {
  return (this.asyncExecutor != null)
      ? CompletableFuture.supplyAsync(supplier, this.asyncExecutor)
      : CompletableFuture.supplyAsync(supplier);
}

private Executor delayed() {
  var millis = this.asyncPollInterval.toMillis();
  return (this.asyncExecutor != null)
      ? CompletableFuture.delayedExecutor(millis, TimeUnit.MILLISECONDS, this.asyncExecutor)
      : CompletableFuture.delayedExecutor(millis, TimeUnit.MILLISECONDS);
}
```

I know this contradicts what I wrote in #664 — my mistake, sorry. The reason: `ForkJoinPool.commonPool()` is *not* identical to `CompletableFuture`'s default executor. `CompletableFuture` uses `ASYNC_POOL`, which is the common pool only when `ForkJoinPool.getCommonPoolParallelism() > 1`, and otherwise a thread-per-task fallback. `supplyAsync(s, commonPool())` is saved by `screenExecutor()`, but `delayedExecutor(delay, unit, commonPool())` on this line is **not** screened — so on a 1-vCPU container (or with `-Djava.util.concurrent.ForkJoinPool.common.parallelism=0`) we'd now push a blocking poll onto a pool with no workers, where the current code falls back to a real thread.

Deferring to `CompletableFuture` when nothing is configured also makes the "when not set, behaviour is unchanged" promise literally true instead of approximately true.

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/DoclingServeClient.java
```diff
@@ -129,12 +132,13 @@ protected DoclingServeClient(DoclingServeClientBuilder builder) {
     this.apiKey = builder.apiKey;
     this.asyncPollInterval = builder.asyncPollInterval;
     this.asyncTimeout = builder.asyncTimeout;
+    this.asyncExecutor = (builder.asyncExecutor != null) ? builder.asyncExecutor : ForkJoinPool.commonPool();
```
edeandrea: Same point as the `AsyncOperations` comment: please drop the `ForkJoinPool.commonPool()` fallback and pass `builder.asyncExecutor` through as-is (`@Nullable`), letting `AsyncOperations` decide between the 1-arg and 2-arg `CompletableFuture` calls. That also removes the `ForkJoinPool` import from this class.

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeApi.java
```diff
@@ -202,6 +204,22 @@ default B prettyPrint() {
...
+     * the task, polling for its status and retrieving its result. If not set, async operations run
+     * on {@link java.util.concurrent.ForkJoinPool#commonPool()}.
+     *
+     * <p>The lifecycle of the executor is owned by the caller: the client never shuts it down.
+     *
+     * @param asyncExecutor the executor to use for async operations (must not be null)
+     * @return this builder instance for method chaining
+     * @throws IllegalArgumentException if asyncExecutor is null
+     */
+    B asyncExecutor(Executor asyncExecutor);
```
edeandrea: Two things on this method:

1. Adding it as `abstract` breaks any third-party implementation of `DoclingApiBuilder` (we expose `DoclingServeApiBuilderFactory` as an SPI, so those exist by design) — source-incompatible at compile time, `AbstractMethodError` at runtime. This was the exact question you raised in #664 and my answer there addressed builder *additivity*, not interface implementors. Please make it a `default` method that throws `UnsupportedOperationException`.
2. Worth documenting the direct-executor case: if a caller passes `Runnable::run`, the initial `supplyAsync` runs the blocking HTTP submit on the calling thread, so `convertSourceAsync` silently becomes partially blocking. A sentence next to the lifecycle note is enough.

docling-serve/docling-serve-client/src/test/java/ai/docling/serve/client/DoclingServeClientAsyncExecutorTests.java
```diff
@@ -0,0 +1,169 @@
...
+        .toCompletableFuture();
+
+    assertThatThrownBy(() -> future.get(10, TimeUnit.SECONDS))
+        .hasRootCauseMessage("Async conversion failed for task %s: Task failed".formatted(TASK_ID));
+
+    // Submitting the task, the first poll and the delayed second poll
+    wireMock.verify(2, getRequestedFor(urlPathEqualTo("/v1/status/poll/" + TASK_ID)));
+    assertThat(this.asyncThreadNames)
+        .isNotEmpty()
+        .allSatisfy(threadName -> assertThat(threadName).startsWith(THREAD_PREFIX));
```
edeandrea: This assertion doesn't prove the thing the PR claims. `asyncThreadNames` is only populated by tasks that actually went through `recordingExecutor`, so "non-empty and all names have the prefix" would still pass if the delayed re-poll — or any other step — silently bypassed the custom executor. That's precisely the regression this test exists to catch.

Please count submissions instead, e.g. an `AtomicInteger` incremented in `recordingExecutor` and asserted to be `>= 4` (submit + poll #1 + delayed re-poll + terminal step), and keep the prefix check as a secondary assertion.

docling-serve/docling-serve-client/src/test/java/ai/docling/serve/client/DoclingServeClientAsyncExecutorTests.java
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  @Test
+  void nullAsyncExecutorIsRejected() {
+    assertThatThrownBy(() -> clientBuilder().asyncExecutor(null))
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessageContaining("asyncExecutor");
+  }
+
+  private static DoclingServeJackson2Client.Builder clientBuilder() {
```
edeandrea: Two coverage gaps:

1. Every scenario here ends in `failure`, so the `SUCCESS` branch — `supplyAsync(() -> getTaskResult(taskResult), asyncExecutor)` — is never exercised, even though the PR description says result retrieval is covered. Please add a scenario that reaches `/v1/result/`.
2. This hardcodes `DoclingServeJackson2Client`, so the Jackson 3 client is untested. Our client tests follow the `AbstractDoclingServeClientTests` + `DoclingServeJackson2ClientTests` / `DoclingServeJackson3ClientTests` pattern — please follow it here too. A `chunkSourceWith*ChunkerAsync` case would be good as well, since `ChunkOperations` gets the executor but has no test at all.

docs/src/doc/docs/whats-new.md
```diff
@@ -25,6 +25,7 @@ Docling Java {{ gradle.project_version }} includes important breaking changes, a
 
 ### {{ gradle.project_version }}
 
+* **Custom `Executor` for async operations** — The async methods (`convertSourceAsync`, `convertSourceBatchAsync`, `convertFilesAsync`, `chunkSourceWith*ChunkerAsync`, ...) used to always run on `ForkJoinPool.commonPool()`. A new `asyncExecutor(Executor)` builder method lets you run them (task submission, status polling and result retrieval) on your own executor instead, e.g. a virtual-thread executor or one managed by your framework. When not set, the behaviour is unchanged. The client never shuts the executor down.
```
edeandrea: Nit, but it's the same inaccuracy that drove the default-executor comment: the async methods didn't "always run on `ForkJoinPool.commonPool()`" — they ran on `CompletableFuture`'s default executor, which is the common pool only when its parallelism is > 1. Suggest: "used to run on `CompletableFuture`'s default executor (usually `ForkJoinPool.commonPool()`)".

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/operations/ChunkOperations.java
```diff
@@ -19,9 +20,8 @@ public final class ChunkOperations extends AsyncOperations implements DoclingSer
   private final HttpOperations httpOperations;
   private final DoclingServeTaskApi taskApi;
 
-  public ChunkOperations(HttpOperations httpOperations, DoclingServeTaskApi taskApi,
-                         Duration asyncPollInterval, Duration asyncTimeout) {
-    super(httpOperations, taskApi, asyncPollInterval, asyncTimeout);
+  public ChunkOperations(HttpOperations httpOperations, DoclingServeTaskApi taskApi, Duration asyncPollInterval, Duration asyncTimeout, Executor asyncExecutor) {
```
edeandrea: This `public` constructor changed signature rather than gaining an overload — that's a source- and binary-incompatible change. Please make the break explicit in the what's-new entry. Also worth adding Javadoc here, including the null/default semantics of `asyncExecutor`, since `ConvertOperations` documents its equivalent.

edeandrea (commented):
Thanks for the work here @Naman-Gururani — the design is what we agreed on in #664 and the plumbing through `AsyncOperations` -> `ConvertOperations` / `ChunkOperations` / `toBuilder()` looks right. A few things I'd like to see addressed before merging.

**One correction to my own advice in #664, and I apologize for the churn:** I suggested defaulting to `ForkJoinPool.commonPool()`. After reviewing the code I think that's the wrong call — please keep the `Executor` `@Nullable` all the way down and only pass it to `supplyAsync` / `delayedExecutor` when it was actually set. Details in the inline comment on `AsyncOperations`.

The rest are smaller: test assertions that don't quite prove what the PR claims, doc nit.

One more thing, not inline: the 4-arg `public` constructors of `ConvertOperations` and `ChunkOperations` were replaced rather than overloaded, which is a source- and binary-incompatible change to published public API. Call it out explicitly in the what's-new entry.

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/operations/AsyncOperations.java
Naman-Gururani: Done in f576a77: the executor stays `@Nullable` all the way down, and `AsyncOperations` goes through `supply(...)` / `delayed()` as suggested, using the 1-arg `supplyAsync` and 2-arg `delayedExecutor` when no executor is set, so the unset case is exactly the previous behaviour.

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/DoclingServeClient.java
Naman-Gururani: Done in f576a77: `builder.asyncExecutor` is passed through as-is and the `ForkJoinPool` import is gone.

docling-serve/docling-serve-api/src/main/java/ai/docling/serve/api/DoclingServeApi.java
Naman-Gururani: Done in f576a77: `asyncExecutor` is now a `default` method throwing `UnsupportedOperationException`, with a test in `DoclingServeApiTests` using a proxy that only implements the abstract methods. The Javadoc covers direct executors. Besides the blocking initial submit, a direct executor would also run the later polls on `CompletableFuture`'s shared delay-scheduler thread, since the delayed executor hands its task to the executor from that thread, so the note mentions both.

docling-serve/docling-serve-client/src/test/java/ai/docling/serve/client/DoclingServeClientAsyncExecutorTests.java
Naman-Gururani: Done in f576a77: the recording executor counts every task submitted to it, and the tests assert the exact count (5 for a successful task: submit, first poll, delay before the re-poll, re-poll, result; 4 for a failed one), keeping the prefix check as a secondary assertion. I used exact counts rather than `>= 4` because on the success path a bypassed delayed re-poll still leaves 4 submissions. With `delayed()` temporarily ignoring the custom executor, 4 of the 6 tests fail.

docling-serve/docling-serve-client/src/test/java/ai/docling/serve/client/DoclingServeClientAsyncExecutorTests.java
Naman-Gururani: Done in f576a77: the tests are now in `AbstractDoclingServeClientAsyncExecutorTests`, run by `DoclingServeJackson2ClientAsyncExecutorTests` and `DoclingServeJackson3ClientAsyncExecutorTests`. The success scenarios go through `/v1/result/`, and there's a `chunkSourceWithHybridChunkerAsync` case. They stay WireMock-only, so they don't need the docling-serve container.

docs/src/doc/docs/whats-new.md
Naman-Gururani: Done in f576a77, with your wording.

docling-serve/docling-serve-client/src/main/java/ai/docling/serve/client/operations/ChunkOperations.java
Naman-Gururani: With the executor nullable again, I restored the previous 4-arg constructors of `ConvertOperations`, `ChunkOperations` and `AsyncOperations` as overloads (delegating with `null`) in f576a77, so the change is additive and there's no break left to call out. All of them have Javadoc now, including the `null` semantics.

docling-serve/docling-serve-api/src/test/java/ai/docling/serve/api/DoclingServeApiTests.java
```diff
@@ -12,6 +15,19 @@ class DoclingServeApiTests {
...
+  @Test
+  void asyncExecutorIsUnsupportedByDefault() {
+    // A builder implementing only the abstract methods of the interface, e.g. one provided through the SPI
+    var builder = (DoclingApiBuilder<?, ?>) Proxy.newProxyInstance(
+        DoclingApiBuilder.class.getClassLoader(), new Class<?>[]{
+            DoclingApiBuilder.class
+        }, (proxy, method, args) -> InvocationHandler.invokeDefault(proxy, method, args));
+
+    assertThatExceptionOfType(UnsupportedOperationException.class)
+        .isThrownBy(() -> builder.asyncExecutor(Runnable::run));
```
edeandrea: Is the `Proxy` needed here? I don't think it is, and I think it actively weakens the test.

The comment above it says "a builder implementing only the abstract methods of the interface" — but `Proxy` generates its implementation at *runtime*, so it can't demonstrate the thing we actually care about from #664: that a third-party builder still **compiles**. That's the whole breaking-change concern.

I tried both versions against a simulated regression (reverting `default` back to `abstract`):

* a plain stub class fails with `error: MinimalBuilder is not abstract and does not override abstract method asyncExecutor(Executor)` — a compile error pointing straight at the cause;
* the `Proxy` version fails with `IllegalArgumentException: "...asyncExecutor(...)" is not a default method` from `Proxy.invokeDefault`, wrapped in an AssertJ "expected UnsupportedOperationException" mismatch. It goes red, but for a reason that reads like a broken test rather than a broken API contract.

Also worth noting: `InvocationHandler.invokeDefault` is a fairly sharp Java 16+ reflective API, and there's no other use of `Proxy` or reflective stand-ins anywhere in the `docling-serve-api` / `docling-serve-client` tests, so this introduces a novel pattern for one test. (Correction to my earlier wording: Mockito *is* available in the project via `quarkus-junit-mockito` — but only in `docling-testing/docling-version-tests`, not in these modules. That doesn't change the argument above, since the issue is that a runtime stand-in can't prove a compile-time guarantee.)

Suggest a plain nested stub instead — I compiled and ran this against the current code and it passes:

```java
@Test
void asyncExecutorIsUnsupportedByDefault() {
  assertThatExceptionOfType(UnsupportedOperationException.class)
      .isThrownBy(() -> new MinimalBuilder().asyncExecutor(Runnable::run));
}

// A builder implementing only the abstract methods of the interface, e.g. one provided through the SPI
private static final class MinimalBuilder implements DoclingApiBuilder<DoclingServeApi, MinimalBuilder> {
  @Override
  public MinimalBuilder baseUrl(URI baseUrl) {
    return this;
  }

  @Override
  public MinimalBuilder apiKey(@Nullable String apiKey) {
    return this;
  }

  // ... logRequests(boolean), logResponses(boolean), prettyPrint(boolean),
  //     connectTimeout, readTimeout, asyncPollInterval, asyncTimeout — all `return this;`

  @Override
  public DoclingServeApi build() {
    throw new UnsupportedOperationException();
  }
}
```

It's more lines, but those lines *are* the test: the class body compiling is the "existing builder implementations keep compiling" guarantee, re-checked on every build. The maintenance cost (updating the stub when an abstract method is added to `DoclingApiBuilder`) is arguably a feature — it forces us to notice the compatibility impact.

docling-serve/docling-serve-client/src/test/java/ai/docling/serve/client/AbstractDoclingServeClientAsyncExecutorTests.java
```diff
@@ -0,0 +1,243 @@
...
+    var response = client(this.recordingExecutor)
+        .convertSourceAsync(convertRequest())
+        .toCompletableFuture()
+        .get(10, TimeUnit.SECONDS);
+
+    assertThat(response).isInstanceOf(InBodyConvertDocumentResponse.class);
+    assertThat(((InBodyConvertDocumentResponse) response).getDocument().getMarkdownContent()).isEqualTo("# Dev");
+
+    // Task submission, first poll, delay before the re-poll, re-poll and result retrieval
+    assertThat(this.executions).hasValue(5);
```
edeandrea: Minor: `hasValue(5)` couples the test to `CompletableFuture` internals. The fifth submission is the `TaskSubmitter` that `CompletableFuture.DelayedExecutor` posts from its scheduler thread — an implementation detail of the JDK, not of this client — so a change to the delayed-execution plumbing would fail this for the wrong reason. The exact count also encodes "exactly one re-poll", which depends on WireMock scenario timing.

`isGreaterThanOrEqualTo(4)` would keep the regression-catching value (a step bypassing the executor still shows up) without the coupling. Happy to leave it exact if you prefer the tighter assertion — just worth a comment noting the 5th is the JDK's delay submission. Same applies at line 133 and line 181.

docling-serve/docling-serve-client/src/test/java/ai/docling/serve/client/AbstractDoclingServeClientAsyncExecutorTests.java
```diff
@@ -0,0 +1,243 @@
...
+    getWireMock().stubFor(get(urlPathEqualTo(RESULT_PATH)).willReturn(okJson(CONVERT_RESULT)));
+
+    var response = client(null)
+        .convertSourceAsync(convertRequest())
+        .toCompletableFuture()
+        .get(10, TimeUnit.SECONDS);
+
+    assertThat(((InBodyConvertDocumentResponse) response).getDocument().getMarkdownContent()).isEqualTo("# Dev");
+    getWireMock().verify(2, getRequestedFor(urlPathEqualTo(POLL_PATH)));
+    getWireMock().verify(1, getRequestedFor(urlPathEqualTo(RESULT_PATH)));
```
edeandrea, line 170: This test verifies the HTTP call counts but no longer asserts that the custom executor was left untouched. Since `recordingExecutor` is built in `setUp()` for every test, `assertThat(this.executions).hasValue(0);` here would cheaply prove the default path never reaches it — right now nothing in the suite would catch an accidental unconditional use of the field.

edeandrea (commented):
Thanks @Naman-Gururani — this addresses everything from the last round, and I verified each point against the source rather than the commit messages.

The default-executor fix is exactly right: `asyncExecutor` is `@Nullable` end to end and the new `supply()` / `delayed()` helpers fall back to the 1-arg `CompletableFuture` overloads, so the unscreened `delayedExecutor(..., commonPool())` hazard is gone. The comment explaining *why* the common pool isn't equivalent to `CompletableFuture`'s default executor is a nice touch. `asyncExecutor` as a `default` method, the restored 4-arg constructor overloads, the `Abstract...` + Jackson2/Jackson3 test split, the submission counter, and the SUCCESS + chunking scenarios all look good.

I ran it locally at this head: 6 tests pass per Jackson client, `DoclingServeApiTests` passes, `spotlessCheck` clean on both modules. I also cross-checked the stubbed paths against the real `TaskOperations` endpoints — they match.

Three minor things left, none blocking. The `Proxy` one is the most substantive.

docling-serve/docling-serve-api/src/test/java/ai/docling/serve/api/DoclingServeApiTests.java
Naman-Gururani: Done in 2832081: replaced the `Proxy` with a `MinimalBuilder` stub that implements only the abstract methods, so a method becoming abstract again now breaks the build rather than the test.

docling-serve/docling-serve-client/src/test/java/ai/docling/serve/client/AbstractDoclingServeClientAsyncExecutorTests.java
Naman-Gururani: Kept them exact and documented them in 2832081. The expected values are now named constants, with a comment listing each submission, including the one `CompletableFuture.delayedExecutor` makes to its base executor after the delay. That submission is part of the documented `delayedExecutor(long, TimeUnit, Executor)` contract ("submits a task to the given base executor after the given delay"). The number of polls is fixed by the scenario states rather than by timing. My reason for keeping them exact: with `isGreaterThanOrEqualTo(4)`, a `delayed()` that bypasses the custom executor still leaves 4 submissions on the success paths, so only the failure test would catch it.

docling-serve/docling-serve-client/src/test/java/ai/docling/serve/client/AbstractDoclingServeClientAsyncExecutorTests.java
Naman-Gururani: Done in 2832081.

edeandrea (approved):
Approving — thank you for this, @Naman-Gururani, and for your patience through several rounds of review.

The final design is exactly right. Keeping `asyncExecutor` `@Nullable` end to end and deferring to `CompletableFuture`'s own default when it isn't set means the "nothing changes unless you set it" promise is literally true, including on low-parallelism hosts where `commonPool()` and `CompletableFuture`'s default executor diverge. You absorbed that redesign after I reversed my own advice from #664, which was my mistake — thanks for taking the rework in stride.

A few things I particularly appreciated:

- **The `MinimalBuilder` stub does real work.** I checked: reverting `default` back to `abstract` on `DoclingApiBuilder#asyncExecutor` now fails at compile time naming the exact method, which is precisely the guarantee we care about for SPI implementors.
- **The named submission-count constants**, with the comment spelling out that the delay step is `delayedExecutor` posting to its base executor once the delay elapses. That's the kind of note that saves the next reader twenty minutes.
- **Following the established test pattern** (`Abstract...` + Jackson2/Jackson3) and covering the chunking path as well as conversion.

Verified locally at 2832081: 14/14 tests pass, `spotlessCheck` clean on both modules, and the stubbed endpoints match the real `TaskOperations` paths.

One footnote, no action needed and not worth another push: `assertThat(this.executions).hasValue(0)` in `asyncOperationsWorkWithoutCustomExecutor` is vacuous — `client(null)` never wires `recordingExecutor` into the client, so the counter is 0 no matter how the client behaves. That one came from my suggestion last round, not from anything you got wrong. I'm leaving it in; noting it here only so nobody later mistakes it for a real guard.

Thanks again for a well-executed contribution.

Naman-Gururani:
Thanks @edeandrea for all the reviews, really appreciate it! Happy to squash the commits before merge if you want.

edeandrea:
> Thanks @edeandrea for all the reviews, really appreciate it! Happy to squash the commits before merge if you want.

If you wouldn't mind that would be great so that the Semantic PR check will pass. I already renamed the PR title accordingly.

Once thats done and all the checks pass I'll go ahead and merge.

Naman-Gururani:
Done, squashed into a single commit with the PR title as the message.

edeandrea:
@all-contributors add @Naman-Gururani for code

edeandrea:
@Naman-Gururani your PR actually got me thinking. Adding additional things to the `DoclingServeApi1 builder is a breaking api change, and the only way around that is the default method solution which throws `UnsupportedOperationException`.

I didn't really like that so I went ahead yesterday and build #696 . I refactored your contribution here to fall into that new model.

Naman-Gururani:
@edeandrea makes sense, the config split is way cleaner than the default-method workaround. Thanks for adding me to the contributors too!

